PR #8, opened 2025-08-08
fix: disable ShrinkWrap for jeandle

taofengliu:
## Related issue(s):

issue #

ShrinkWrap moves the epilog and prolog away from the functino's entry and exit. This shift can result in issues where rbp, used temporarily to store the return value, is overwritten by the pop instruction in epilog, corrupting the return value.


Cruise20 (approved):

taofengliu:
Need a test case for this fix. I will add it soon.

Issue #14, opened 2025-08-16
Support AArch64


feilongjiang:
**The feature you'd like**

Add support for AArch64.

**Implementation you've considered**

Add Hotspot Calling convention for aarch64.

**Additional context**

PR #15, opened 2025-08-16
feat: Initial support for AArch64

feilongjiang:
## Related issue(s):

issue #14 

WIP

Hi team, please consider.
This PR contains initial support for the linux-aarch64 platform. The major changes are:
1. Add a new Java calling convention named `CC_AArch64_Hotspot`, including function args assignment convention and callee-saved register list.
2. Even if we set `FP` and `LR` as callee-saved registers, `TargetFrameLowering::determineCalleeSaves` may treat `FP` as a non-saved register when `MRI.isPhysRegModified(Reg)` returns `false` (which means `FP` was not modified across the calls, maybe?). To make stack unwinding correct, we should save `FP` manually.

Testing:
- [x] jeandle-jdk tier1_compiler_jeandle with linux-aarch64 slowdebug build
- [ ] jeandle-jdk tier1_compiler_jeandle with linux-aarch64 release build

### DIFF
--- a/llvm/lib/Target/AArch64/AArch64CallingConvention.td
+++ b/llvm/lib/Target/AArch64/AArch64CallingConvention.td
@@ -553,6 +553,40 @@ def CC_AArch64_Preserve_None : CallingConv<[
   CCDelegateTo<CC_AArch64_AAPCS>
 ]>;
 
+//===----------------------------------------------------------------------===//
+// ARM64 Calling Convention for OpenJDK HotSpot
+//===----------------------------------------------------------------------===//
+
+// This calling convention is specific to the OpenJDK HotSpot JIT Compiler.
+// The only documentation is the OpenJDK source code, specifically the C header
+// file:
+//
+//    https://github.com/jeandle/jeandle-jdk/blob/main/src/hotspot/cpu/aarch64/assembler_aarch64.hpp
+//
+// which defines the registers for the Java calling convention.
+//
+//  |--------------------------------------------------------------------|
+//  | X0       X1       X2      X3      X4      X5      X6      X7       |
+//  |--------------------------------------------------------------------|
+//  | c_rarg0  c_rarg1  c_rarg2 c_rarg3 c_rarg4 c_rarg5 c_rarg6 c_rarg7  |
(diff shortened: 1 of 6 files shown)


.github/workflows/jeandle-llvm-test.yml
```diff
@@ -1,15 +1,30 @@
 name: "jeandle-llvm check"
 
 on:
-  pull_request_target:
-    branches:
-      - 'main'
+  push:
```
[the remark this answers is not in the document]
taofengliu: I think we'd better trigger actions through pull requests.
People always create a PR immediately after pushing their code. And we cannot ignore the main branch because we cannot prohibit contributors from developing on their main branches. But if we don't ignore the main branch, after every merge there are always some meaningless CI actions added to our actions history.

.github/workflows/jeandle-llvm-test.yml
```diff
@@ -1,15 +1,30 @@
...
+    branches-ignore:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
taofengliu, line 9: If we trigger CI actions through PRs, we should write the config like this:
```suggestion
  group: ${{ github.workflow }}-${{ github.head_ref }}
```

llvm/lib/Target/AArch64/AArch64CallingConvention.td
```diff
@@ -691,6 +725,12 @@ def CSR_AArch64_StackProbe_Windows
                            (sequence "X%u", 18, 28), FP, SP,
                            (sequence "Q%u", 0, 31))>;
 
+// HotSpot C1/C2 compiler don't use any callee save registers.
+//   X0-X7,X10-X26 volatile (caller save)
+//   X27-X32 system (no save, no allocate)
+//   X8-X9 non-allocatable (so we can use them as scratch regs)
+def CSR_AArch64_Hotspot : CalleeSavedRegs<(add LR, FP)>;
```
taofengliu, line 732: Using ```OtherPreserve``` maybe a good choice here.

taofengliu (commented):
Some minor comments.

PR #24, opened 2025-09-04
Refactor java operation lower

lfeng14:
---------

## Related issue(s):

issue #

Inlining transitivity and Recursion handling


llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,81 @@
...
+  return V == Phase;
+};
+
+static bool runImpl(Module &M, int Phase, ModuleAnalysisManager &MAM,
+                    FunctionAnalysisManager *FAM) {
+
+  SmallSetVector<CallBase *, 16> Calls;
   bool Changed = false;
-  InlineFunctionInfo IFI;
+  SmallVector<Function *, 16> InlinedComdatFunctions;
```
taofengliu, line 48: Is this used?

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,81 @@
...
-        assert(!Failed && "wrong value of LowerPhase attribute");
-        if (LowerPhaseValue != Phase)
-          continue;
-
-        CallInstructions.push_back(CB);
+    Calls.clear();
+    for (User *U : F.users())
+      if (auto *CB = dyn_cast<CallBase>(U)) {
+        if (CB->getCalledFunction() == &F &&
+            !CB->getAttributes().hasFnAttr(Attribute::NoInline))
```
taofengliu, line 62: Every JavaOp will has the attriburte ```noinline``` to avoid be inlined by other inline passes. We don't need to check this.

taofengliu:
The test is complex, try to simplify and draw a call graph in comment.

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
lfeng14: Not used, has been deleted

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
lfeng14: The condition has been removed.

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,81 @@
...
-  for (Function &F : M) {
-    LLVM_DEBUG(dbgs() << "Searching in function: " << F.getName()
-                      << " in lower phase: " << Phase << "\n");
-    if (F.isDeclaration())
+  for (Function &F : make_early_inc_range(M)) {
+    if (F.isPresplitCoroutine())
+      continue;
+
+    if (!isPhaseFunc(F, Phase) || F.isDeclaration() ||
+        !isInlineViable(F).isSuccess())
```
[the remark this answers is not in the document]
lfeng14: done

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,82 @@
...
+    return false;
+  int V = 0;
+  bool Failed = F.getFnAttribute(jeandle::Attribute::LowerPhase)
+                    .getValueAsString()
+                    .getAsInteger(10, V);
+  assert(!Failed && "wrong value of LowerPhase attribute");
+  return V == Phase;
+};
+
+static bool runImpl(Module &M, int Phase, ModuleAnalysisManager &MAM,
```
taofengliu, line 43: MAM is not used.

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,82 @@
...
+#include <utility>
 
 using namespace llvm;
 
 #define DEBUG_TYPE "java-operation-lower"
 
 namespace {
 
-static bool runImpl(Module &M, int Phase) {
+bool isPhaseFunc(const Function &F, int Phase) {
```
taofengliu, line 32:
```suggestion
bool shouldInline(const Function &F, int Phase) {
```
I think it's better to define this as a lambda expression.

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,82 @@
...
-        int LowerPhaseValue;
-        bool Failed =
-            LowerPhase.getValueAsString().getAsInteger(10, LowerPhaseValue);
-        assert(!Failed && "wrong value of LowerPhase attribute");
-        if (LowerPhaseValue != Phase)
-          continue;
-
-        CallInstructions.push_back(CB);
+    if (!isPhaseFunc(F, Phase) || F.isDeclaration())
+      continue;
```
taofengliu, line 57: JavaOps cannot be PresplitCoroutine and declaration, so use assert here.

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,82 @@
...
-      if (InlineFunction(*CB, IFI).isSuccess()) {
-        Changed = true;
-        LLVM_DEBUG(dbgs() << "Successfully inlined: " << Callee->getName()
-                          << " in lower phase: " << Phase << "\n");
-      } else {
-        LLVM_DEBUG(dbgs() << "Failed to inline: " << Callee->getName()
+    for (CallBase *CB : Calls) {
+      Function *Caller = CB->getCaller();
+      InlineFunctionInfo IFI;
+      InlineResult Res = InlineFunction(*CB, IFI);
```
taofengliu, line 70: Why we don't invoke InlineFunction like AlwaysInliner does?

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,82 @@
...
+      if (!Res.isSuccess()) {
+        LLVM_DEBUG(dbgs() << "failed to inline: " << Caller->getName()
                           << " in lower phase: " << Phase << "\n");
+        continue;
       }
+
```
taofengliu, line 76: Should we invoke emitInlinedIntoBasedOnCost here?

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,82 @@
...
+      Changed = true;
+      if (FAM)
+        FAM->invalidate(*Caller, PreservedAnalyses::none());
     }
+
+    if (FAM)
+      FAM->clear(F, F.getName());
+    LLVM_DEBUG(dbgs() << "remove unused function: " << F.getName()
+                      << " in lower phase: " << Phase << "\n");
+    M.getFunctionList().erase(F);
```
taofengliu, line 86: AlwaysInliner only do this for InlinedComdatFunctions, do we need this?

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
```diff
@@ -9,65 +9,82 @@
...
+    Changed = true;
```
taofengliu, line 87: This shouldn't be here.

taofengliu (changes requested):

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
lfeng14: Yes, comdat functions also need to undergo this process; the only difference is that comdat functions should use filterDeadComdatFunctions to filter out the truly dead ones.

```

  if (!InlinedComdatFunctions.empty()) {
    // Now we just have the comdat functions. Filter out the ones whose comdats
    // are not actually dead.
    filterDeadComdatFunctions(InlinedComdatFunctions);
    // The remaining functions are actually dead.
    for (Function *F : InlinedComdatFunctions) {
      if (FAM)
        FAM->clear(*F, F->getName());
      M.getFunctionList().erase(F);
      Changed = true;
    }
  }
```

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
lfeng14: No, emitInlinedIntoBasedOnCost is just a diagnostic hook in LLVM's inlining mechanism. It emits an optimization remark recording the cost to help developers trace and understand inlining decisions, but here I'm using alwaysinline.

llvm/test/Jeandle/java-operation-lower.ll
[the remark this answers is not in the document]
lfeng14: So we need to keep the specific phase function and add the 'noline' attribute ？

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
lfeng14: done

llvm/lib/Transforms/Jeandle/JavaOperationLower.cpp
lfeng14: done

PR #27, opened 2025-10-11
ci: Support for RISC-V and running CI according to labels

lgxbslgx:
This patch added the CI step about RISC-V and run CI according to labels of the PR.


.github/workflows/jeandle-llvm-test.yml
```diff
@@ -10,35 +11,28 @@ concurrency:
...
+      target: AArch64
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'AArch64') }}
+
+  test-linux-riscv64:
+    name: linux-riscv64
+    uses: ./.github/workflows/test-linux.yml
+    with:
+      runs-on: ubuntu-22.04
+      target: RISCV
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'RISC-V') }}
```
taofengliu: LLVM can configure more than one target in one compilation, so we can merge separate test jobs in one. Just use labels to determine which targets we need to include.

PR #28, opened 2025-10-11
ci: enable a workflow to be triggered manually

lgxbslgx:

Currently, we can only trigger a workflow by submitting a PR. It is good to enable a workflow to be triggered manually in the `actions` tab.


lgxbslgx:
OK, closing this PR.

PR #59, opened 2026-04-13
feat: Type query interface and type check elimination

taofengliu:
## Related issue(s):

https://github.com/jeandle/jeandle-jdk/issues/370


llvm/lib/Transforms/Jeandle/TypeCheckElimination.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  bool Changed = false;
+  for (CallInst *CI : Checks) {
+    uintptr_t SuperKlass = jeandle::extractKlassConstant(CI->getArgOperand(0));
+    if (SuperKlass == 0)
+      continue;
+
+    Value *Obj = CI->getArgOperand(1);
+    jeandle::JavaType ObjType = jeandle::getJavaType(Obj, DT, CI);
+
```
minchao1024: The current dominator tree traversal only merges results where the `instanceof` check succeeds. Maybe we should also capture the `false` results.
```
// TestTypeAnalysis.java
// java -Xcomp -XX:-TieredCompilation -XX:CompileThreshold=10 -XX:CompileCommand=compileonly,TestTypeAnalysis::test -XX:CompileCommand=printassembly,TestTypeAnalysis::test TestTypeAnalysis
class TestTypeAnalysis {
    public static void test(Object obj) {
        if (obj instanceof Exception) {
            System.out.println("assert success!");
            if (obj instanceof Exception) { // Replaced with true now
                blaceHole();
            }
        } else {
            if (obj instanceof Exception) { // Not replaced with false yet
                System.out.println("assert fail!");
            }
        }
    }
    public static void blaceHole() {
    }
    public static void main(String[] args) throws Throwable {
        for (int i = 0; i < 15; i++) {
            if (i == 10) {
                test(new Object());
            } else {
                test(new RuntimeException());
            }
        }
        Thread.sleep(1000);
    }
}
```

llvm/lib/IR/Jeandle/JavaType.cpp
```diff
@@ -0,0 +1,591 @@
...
+  assert(CB && CB->IsSubtype && "VMCallbacks must be set");
+
+  BasicBlock *ContextBB = Context->getParent();
+  JavaType Best;
+
+  for (auto *Node = DT.getNode(ContextBB); Node; Node = Node->getIDom()) {
+    BasicBlock *BB = Node->getBlock();
+    auto *BI = dyn_cast<BranchInst>(BB->getTerminator());
+    if (!BI || !BI->isConditional())
+      continue;
```
minchao1024, line 952: In the first iteration, `BB` is `ContextBB`. If `Context` appears before the terminator `BI`, the branch decision cannot provide any type information for the `Context` instruction. Thus, we can skip the first iteration in this situation.
```suggestion
  for (auto *Node = DT.getNode(ContextBB); Node; Node = Node->getIDom()) {
    BasicBlock *BB = Node->getBlock();
    auto *BI = dyn_cast<BranchInst>(BB->getTerminator());
    if (BB == ContextBB && BI != Context)
      continue;
    if (!BI || !BI->isConditional())
      continue;
```

llvm/lib/Transforms/Jeandle/TypeCheckElimination.cpp
taofengliu: I've support this now. But it's a time consuming operation. Maybe we need to pay attention to the trade-off.

llvm/lib/IR/Jeandle/JavaType.cpp
taofengliu: We can just skip the current block.
```
for (auto *Node = DT.getNode(ContextBB)->getIDom(); Node; Node = Node->getIDom()) {
    BasicBlock *BB = Node->getBlock();
    auto *BI = dyn_cast<BranchInst>(BB->getTerminator());
    if (!BI || !BI->isConditional())
      continue;
```

minchao1024: We cannot simply skip the current block. because during Phi node processing, we must merge the types from all incoming edges. In this process, we use the terminator of the incoming block as the context to recursively get type information. 

This leads to a logical inconsistency: 
- for non-Phi nodes, the context node is the node about to be executed, which does not affect current type analysis;
- however, for Phi nodes, each context node represents a node that has already been executed, so we need to think about its impact on type sharpening.
```
    public static void testPhi(Object obj) {
        obj = new String();
        while (obj instanceof String) {  // This incoming block helps
            if (obj instanceof String) { // Handle phi node here
                blaceHole();
            }
            obj = new Object();
        }
    }
```

taofengliu: In fact, even we don't skip the current block, the process of phi nodes is wrong. If the terminators of incoming blocks are conditional branches from type checks, we cannot use the type checks' information. Because in the implementation, we use `DT.dominates(TypeConfirmedBB, ContextBB)`, the TypeConfirmedBB is the phi node's block, the ContextBB is the incoming block. `DT.dominates(TypeConfirmedBB, ContextBB)` will never be true.
I have fixed it.

minchao1024 (approved):

Issue #62, opened 2026-04-28
Support test Jeandle's LLVM passes separately from JDK.


taofengliu:
**The feature you'd like**
We have VM callbacks for Jeandle-LLVM now, which can provide JVM runtime information for passes. But it makes it impossible to test Jeandle-LLVM without JVM.
Now we want to add comprehensive tests for Jeandle-LLVM, making it cover the corner cases that are hard to construct from Java code. So a mechanism to test Jeandle-LLVM separately is necessary.

**Implementation you've considered**
Maybe we can dump the callbacks' return values to a file, then run Jeandle-LLVM's tests with this file.

PR #66, opened 2026-05-08
feat: support vm callback dump/replay

taofengliu:
## Related issue(s):

issue #62

Note: All tests about type check eliminations are just for VM callback dump/replay. I will add more comprehensive tests for type check elimination in the next PR.


taofengliu:
The format check fails because the ALL_JEANDLE_VM_CALLBACKS macro. But we need to declare all VM callbacks like this, so just bypass the test.

llvm/include/llvm/IR/Jeandle/VMCallback.h
```diff
@@ -21,38 +21,87 @@
...
+//   Params   — parenthesized parameter declarations,
+//              e.g. (uintptr_t a1, uintptr_t a2)
+//   Args     — parenthesized argument names, e.g. (a1, a2)
+//   ArgTypes — parenthesized VMCallbackValueType enum values, e.g.
+//              (VMCallbackValueType::Uintptr, VMCallbackValueType::Uintptr)
+//   NumArgs  — number of arguments
+//
+// To add a new callback, add one row below, then implement the JDK-side
+// function in jeandleVMCallback.cpp and wire it in
+// register_jeandle_vm_callbacks().
```
minchao1024, line 70: We should add a comment to highlight the VM callback Sequential Consistency Requirement
```suggestion
// ALL_JEANDLE_VM_CALLBACKS(def) invokes `def` for each VM callback with:
//   Name     — callback name (struct field, CK_ prefix, stringification)
//   RetType  — C++ return type (bool, uintptr_t)
//   ResType  — VMCallbackValueType enum suffix (Bool, Uintptr, Int)
//   Params   — parenthesized parameter declarations,
//              e.g. (uintptr_t a1, uintptr_t a2)
//   Args     — parenthesized argument names, e.g. (a1, a2)
//   ArgTypes — parenthesized VMCallbackValueType enum values, e.g.
//              (VMCallbackValueType::Uintptr, VMCallbackValueType::Uintptr)
//   NumArgs  — number of arguments
//
// -----------------------------------------------------------------------------
// CRITICAL: Sequential Consistency Requirement
//
// The VMCallback Replay mechanism relies on a strict total ordering of callback 
// invocations. For a replay to be successful, the sequence of VM calls during 
// compilation must be identical to the sequence recorded in the log.
//
// Developers must ensure that any logic triggering VM callbacks follows a 
// deterministic execution path. Avoid issuing callbacks while iterating over 
// unordered containers (e.g., llvm::DenseMap or std::unordered_map). 
//
// Correct Example:
//   for (Instruction &I : instructions(F)) { triggerCallback(I); }
//
// Incorrect Example:
//   for (auto &Entry : UnorderedMap) { triggerCallback(Entry.second); }
// -----------------------------------------------------------------------------
//
// To add a new callback, add one row below, then implement the JDK-side
// function in jeandleVMCallback.cpp and wire it in
// register_jeandle_vm_callbacks().
```

llvm/lib/IR/Jeandle/VMCallbackLog.cpp
```diff
@@ -0,0 +1,379 @@
...
+// REPLAY_CALLBACK(Name, RetType, (param-decls), (arg-names))
+// Same parenthesized-params pattern as RECORD_CALLBACK.
+#define REPLAY_CALLBACK(Name, RetType, Params, Args)                           \
+  static RetType replay##Name Params {                                         \
+    if (consumeEntry(CK_##Name, encodeArgs(JEANDLE_STRIP_PARENS Args), #Name)) \
+      return decodeVMCallbackValue<RetType>(                                   \
+          LogData->Entries[LogData->Cursor - 1].Result);                       \
+    return decodeVMCallbackValue<RetType>(                                     \
+        defaultResult(getCallbackTable()[CK_##Name].ResType));                 \
+  }
```
minchao1024, line 224: Replay is a debugging tool that relies on strict determinism. If `consumeEntry` returns false, it means the execution path has deviated from the log. We should immediately terminate rather than returning a defaultResult.

taofengliu: done

llvm/include/llvm/IR/Jeandle/VMCallback.h
taofengliu: done

minchao1024 (approved):

Issue #68, opened 2026-05-26
Add Jeandle-specific inlining pass for Java methods


minchao1024:
**The feature you'd like**

A custom inlining pass for Jeandle that can inline Java method calls where the callee may initially be a declaration.

**Implementation you've considered**

A Module-level pass that walks call sites directly, delegates inlining decisions and callee resolution to VMCallbacks, and supports multi-level inlining with depth control and recursion prevention.

PR #69, opened 2026-05-26
feat: Add JVM-guided Java method inlining

minchao1024:
## Related issue(s):

issue #68 

Add a Jeandle inline driver and inliner for JVM-guided Java method
inlining, with callback replay support and focused lit coverage.

This change:
- Adds JeandleInlineDriver and JeandleInliner to inline monomorphic Java
  method call sites under JVM policy control.
- Queries VM callbacks for inline policy, callee IR materialization, inline
  success notification, and inline-phase completion.
- Tracks inline scope IDs so nested inline decisions can be replayed with the
  correct caller scope.
- Supports replaying GetInlineCalleeIR from companion *_inline_callees.ll
  modules during opt callback-log replay.
- Materializes replay callee definitions on demand, including referenced
  globals, aliases, helper declarations, and EH personality wiring.
- Wires the inline driver into the Jeandle pipeline and registers
  jeandle-inline-driver for opt tests.
- Adds lit tests for global variables, global aliases, exception handling,
  policy rejection, and resume handling.


llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
```diff
@@ -0,0 +1,263 @@
...
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements the JeandleInliner, a Module-level inlining pass
+// for Jeandle JVM JIT. Unlike the standard InlinerPass which operates on the
+// LazyCallGraph/SCC, this pass walks all call sites in the module directly.
+//
+// Algorithm:
+//   1. Collect all call sites where the callee is a Jeandle Java method
+//      (identified by F.hasGC() && F.getGC() == jeandle::JeandleGC).
```
taofengliu: We can attach "java-method" attribute to Java method declarations, making them more clear.

llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
```diff
@@ -0,0 +1,221 @@
...
+    // exposed by the inline round. Do not carry a cross-step worklist through
+    // this point; the next inline round rescans the root function and should
+    // read the preserved inline-scope-id metadata from surviving/generated call
+    // sites.
+    if (!DevirtResult.AddedMonomorphicTargets)
+      break;
+  }
+
+  if (!Changed)
+    return PreservedAnalyses::all();
```
taofengliu, line 244:
```suggestion
  if (DriverPa.areAllPreserved())
    return PreservedAnalyses::all();
```
Then we can delete the `Changed` variable. Then JeandleInliner and JeandleDevirtualization don't need to return `Changed` in their results.

llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
```diff
@@ -0,0 +1,910 @@
...
+
+      Function *NewCallee = NewCB->getCalledFunction();
+      // TODO: Support inlining the root method as a callee once root/caller IR
+      // and callee IR use the same unwind model. The root emits real unwind
+      // operations, while an inlined callee forwards unwind edges to the
+      // caller's landingpad. Mark root callees noinline so later LLVM inline
+      // passes cannot inline them either.
+      if (NewCallee == RootFunction) {
+        NewCB->setIsNoInline();
+        Changed = true;
```
taofengliu: Maybe not needed

llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
```diff
@@ -0,0 +1,910 @@
...
+static PreservedAnalyses getInlineRoundPreservedAnalyses(bool Changed) {
+  if (!Changed)
+    return PreservedAnalyses::all();
+
+  PreservedAnalyses PA;
+  // InlineFunction mutates individual callers and those function analyses are
+  // invalidated eagerly after each inline. Preserve the function-analysis proxy
+  // here so the module pass manager does not discard unrelated function
+  // analyses.
+  PA.preserveSet<AllAnalysesOn<Function>>();
```
taofengliu, line 148: As the comment says, we need to preserve the `FunctionAnalysisManagerModuleProxy`

llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
```diff
@@ -0,0 +1,221 @@
...
+  // Callee IR requested from the JVM is available_externally: it is useful for
+  // optimization, but should not remain as a definition after the inline driver
+  // is done. If such a method still has uses, delete only its body so existing
+  // references stay valid as declarations; otherwise erase the function.
+  bool RemovedCalleeIR = eraseAvailableExternallyJavaMethods(M, RootFunction);
+
+  if (ErasedInlineScopeIDs || RemovedCalleeIR) {
+    PreservedAnalyses CleanupPA = PreservedAnalyses::none();
+    updateDriverPreservedAnalyses(M, MAM, DriverPA, std::move(CleanupPA));
+  }
```
taofengliu: I think a finer-grained control over the analysis would be better, as we already do so in the main inliner loop, don't give up all the maintained analysis. We can only invalidate the effected analysis here.

llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
```diff
@@ -0,0 +1,910 @@
...
+      // The JVM frontend guarantees that callee allocas are static allocas, so
+      // InlineFunction should hoist them into the caller entry block. Recheck
+      // that invariant in debug builds.
+      assertNoNonEntryAlloca(*Caller);
+#endif
+      bool Recorded = VC->RecordInlineSuccess(InlineScopeID, BCI, CalleeMethod);
+      if (!Recorded) {
+        logInlineEvent("record-fail", ScopeCaller, BCI, Callee, InlineScopeID,
+                       ThreadID, InlineScopes);
+      }
```
taofengliu: Here we should make sure `Recorded` is true. If it is false, this would cause an InlineScope to be created on the LLVM side in the code below, without a corresponding InlineTree on the JVM side.

Maybe we should stop inlining in this compilation if `Recorded` is false.

llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
```diff
@@ -0,0 +1,910 @@
...
+    }
+
+    if (Callee->isDeclaration()) {
+      logInlineEvent("request-ir", ScopeCaller, BCI, Callee, InlineScopeID,
+                     ThreadID, InlineScopes);
+      bool GotCalleeIR = VC->GetInlineCalleeIR(CalleeMethod);
+      if (!GotCalleeIR) {
+        logInlineEvent("missing-ir", ScopeCaller, BCI, Callee, InlineScopeID,
+                       ThreadID, InlineScopes);
+        continue;
```
taofengliu, line 489: Here we may leave residual IR from a failed callee compilation in the Module, which needs to be cleaned up/removed. We can remove it in the JDK or LLVM side.

llvm/lib/Jeandle/Pipeline.cpp
```diff
@@ -33,13 +35,27 @@ Pipeline::Pipeline(OptimizationLevel level, LLVMContext &Ctx)
...
+  switch (Options.Inlining) {
+  case InlineMode::Disabled:
+    break;
+  case InlineMode::Default:
+    PM.addPass(JeandleInlineDriver());
+    break;
+  case InlineMode::AccessorOnly:
+    PM.addPass(JeandleInlineDriver(/*InlineAccessorsOnly=*/true));
+    break;
+  }
```
taofengliu, line 58: Would it be better to run some IR simplification passes before inlining, and also between each inlining round?

llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
minchao1024: Done

llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
minchao1024: Done

llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
minchao1024: Fixed

llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
minchao1024: Fixed

llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
minchao1024: Fixed

llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
minchao1024: This record operation is checked and must succeed for now. The only expected failure would be failing to create the inline tree, so a TODO is left to pre-allocate the inline tree before LLVM performs inlining and make this step non-failing in a follow-up change.

llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
minchao1024: Fixed

llvm/lib/Jeandle/Pipeline.cpp
minchao1024: Done; `InstSimplifyPass` has now been added between inlining rounds. This step was not included prior to inlining because the potential for optimization appeared limited.

llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
```diff
@@ -0,0 +1,270 @@
...
+  // Loop shape:
+  //   1. Run one inline round. The round tags every newly exposed call site
+  //      with inline-scope-id metadata.
+  //   2. If the inline round did not expose any new call sites, stop.
+  //   3. Run devirtualization refinement. It must propagate inline-scope-id
+  //      and deopt/BCI information when it clones or replaces calls.
+  //   4. If devirtualization preserved everything, it did not produce a new
+  //      MonomorphicTarget call site and the loop stops; otherwise rescan IR
+  //      in the next inline round.
+  for (;;) {
```
taofengliu: Maybe it's better to have a `MaxIterations` cap

llvm/lib/Jeandle/Pipeline.cpp
```diff
@@ -63,9 +63,17 @@ ModulePassManager Pipeline::buildJeandlePipeline(PassBuilder &PB,
...
   PM.addPass(RewriteStatepointsForGC());
+  // Phase 9 is reserved for JavaOps that must be lowered after RS4GC. In
+  // particular, G1 barriers can materialize raw addresses derived from oops
+  // (for example card-table addresses). If those addresses are created before
+  // RS4GC and optimized across safepoints, they are not relocated when the
+  // source oop moves. Lower them after RS4GC so the optimizer cannot reuse
+  // stale raw derived addresses across safepoints.
+  PM.addPass(JavaOperationLower(9));
+  PM.addPass(createModuleToFunctionPassAdaptor(TLSPointerRewrite()));
+  PM.addPass(createModuleToFunctionPassAdaptor(InstSimplifyPass()));
```
taofengliu: Since GC barriers aren't inlined until O3, we might as well delay inserting them until after O3, otherwise they'll get in the way of optimizations.

llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
minchao1024: Done. Added `MaxInlineDriverIterations` as a guard for abnormally long inline loops. It does not affect the normal inlining path.

llvm/lib/Jeandle/Pipeline.cpp
minchao1024: I think it is better to keep InsertGCBarriers before O3.

Running InsertGCBarriers before O3 still lets O3 optimize the high-level barrier form, and that does not affect correctness. The correctness issue only appears after barrier lowering, where G1 barriers create raw intermediate values derived from oops.If those lowered intermediate values are exposed before O3, O3 may move or reuse them across safepoints. That is the unsafe part, because those raw values are not updated when the oop moves.

So the split is intentional: insert barriers before O3 for optimization, but lower them after O3/RS4GC to keep those raw intermediate values from crossing safepoints.

taofengliu: The GCBarriers may affect optimizations, leave a TODO here to note it.

minchao1024: Done

llvm/lib/Jeandle/Pipeline.cpp
```diff
@@ -33,23 +35,59 @@ Pipeline::Pipeline(OptimizationLevel level, LLVMContext &Ctx)
...
   PM.addPass(createModuleToFunctionPassAdaptor(InstSimplifyPass()));
   PM.addPass(createModuleToFunctionPassAdaptor(TypeCheckElimination()));
   PM.addPass(createModuleToFunctionPassAdaptor(RepeatedConstantFolding()));
   PM.addPass(createModuleToFunctionPassAdaptor(TypeCheckElimination()));
+  // TODO: InsertGCBarriers currently inserts high-level barrier calls before
+  // O3. This is a conservative placeholder to keep Java heap stores
+  // GC-observable during optimization, but the uninlined barrier calls can
+  // still block useful optimizations. Consider a better representation with
+  // more precise memory effects or another scheme that preserves GC semantics
+  // without exposing heavy calls to the main optimizer.
```
taofengliu:
```suggestion
  // TODO: InsertGCBarriers currently inserts high-level barrier calls before
  // O3 because it cannot handle O3 generated memory intrinsics and vector 
  // instructions. But the uninlined barrier calls can still block useful 
  // optimizations.
```

taofengliu (approved):

Issue #52, opened 2026-02-27
Eliminate safepoint in counted loop


kuaiwei:
**Goals**
Safepoint in loop block other optimization like auto vectorization. C2 and Graal will remove safepoint inside counted loop. Jeandle could do same optimization

PR #70, opened 2026-05-26
enhance: Eliminate safepoint in counted loop - Phase 1

eapenzhan:
## Related issue(s):

issue #52 

This PR adds Phase 1 of Jeandle safepoint poll elimination.

Jeandle currently emits `jeandle.safepoint_poll` calls on loop back-edge paths. These opaque calls block several downstream LLVM optimizations, including LICM, LoopVectorize, loop unrolling, and closed-form simplification. This change introduces a conservative LLVM loop pass that removes those polls only when the loop trip count is provably bounded.

## Design

The new `SafepointElimination` pass is implemented as a `LoopPass` and runs through `FunctionToLoopPassAdaptor`. For each loop, it:

- acts only on Jeandle Java method compilations, identified by the existing module metadata;
- collects `jeandle.safepoint_poll` calls whose blocks dominate the loop latch, i.e. polls on the loop back-edge path;
- classifies a loop as bounded using `ScalarEvolution::getConstantMaxBackedgeTakenCount`;
- removes only back-edge polls in bounded loops;
- preserves polls on side-exit or deoptimization paths, such as `uncommon_trap` slow paths;
- leaves unbounded loops unchanged.

This mirrors the first-stage behavior of HotSpot C2 counted-loop safepoint stripping and Graal's loop safepoint elimination: bounded inner loops can become poll-free, while loops that may run for a long time still keep their safepoint poll until a future strip-mining phase is added.

The pass is wired into the Jeandle pipeline after Java operation lowering and an `EarlyCSE` pass, but before the default LLVM optimization pipeline. `EarlyCSE` helps expose loop-invariant bounds, for example by hoisting calls such as `jeandle.arraylength`, so that SCEV can prove common Java array loops bounded before vectorization and other loop optimizations run.

## Safety

The implementation is intentionally conservative:

- the master switch `-jeandle-enable-safepoint-elim` can disable the pass;
- `-jeandle-short-loop-max-iter` controls the short-loop threshold;
- the Graal-style 32-bit range rule is available behind `-jeandle-sp-elim-32bit-range`, but defaults to false until strip-mining is available;
- unbounded loops keep their back-edge safepoint polls;
- side-exit polls carrying deoptimization state are preserved.

## Tests

Added lit coverage for the positive and negative Phase A cases:

- simple counted loops;
- non-zero initial values;
- negative and non-unit strides;
- bounded i64 loops;
- multiple safepoints in one loop;
- nested counted loops;
- realistic Jeandle-generated array-loop IR;
- non-counted loops;
- bad `icmp ne` forms that SCEV cannot safely bound;
- side-exit poll preservation;
- non-Jeandle/non-Java modules;
- default-off behavior for the 32-bit-range rule.

## Follow-up Work

This PR intentionally implements only the first, conservative phase of safepoint elimination. The following work is left for subsequent changes:

- Add loop strip-mining for long-running or otherwise unbounded loops, similar to HotSpot C2's `OuterStripMinedLoopNode`, so that those loops can still expose poll-free inner bodies while keeping bounded GC latency.
- Revisit `-jeandle-sp-elim-32bit-range` once strip-mining exists. At that point, the 32-bit-range rule can either become safe by construction, be routed through strip-mining, or be removed if it is no longer needed.
- Add broader Jeandle JIT regression tests that check generated IR or optimization remarks from real Java loops, rather than only standalone LLVM IR fixtures.
- Add performance measurements for representative Java array loops and allocation-heavy kernels to quantify the optimization impact on LICM, vectorization, and loop unrolling.
- Extend the analysis, if needed, to recognize more loop shapes that SCEV can prove finite but that are not yet covered by the current fixtures.


llvm/include/llvm/Transforms/Jeandle/SafepointElimination.h
```diff
@@ -0,0 +1,38 @@
...
+class LPMUpdater;
+class Loop;
+
+/// LoopPass that strips `jeandle.safepoint_poll` calls out of the back-edge
+/// path of counted loops whose trip count we can prove fits within the
+/// configured budget. Driven via createFunctionToLoopPassAdaptor so that
+/// LoopSimplify and LCSSA run before us, and so a future strip-mining
+/// transform can register newly-created outer loops with the LPM via
+/// LPMUpdater.
+class SafepointElimination : public PassInfoMixin<SafepointElimination> {
```
taofengliu, line 28: Don't implement it as a loop pass, because we will extend it to eliminate more safepoints not only in loop scopes.

PR #71, opened 2026-05-29
fix: stop Reassociate sinking loop-carried recurrence phis

yadongw:
## Problem

Reassociate ranks a loop header phi at the loop's lowest BB-RPO position, so the
"low rank goes to the RHS" canonicalization treats a recurrence accumulator like
a loop invariant and sinks it into the innermost operand of the add chain. The
loop-carried critical path goes from one ALU op to several:

```
s += a*7 + (a>>2)
  before:  ((a>>2) + s) + a*7     ; 3-op loop carry on s
  after:   (a*7 + (a>>2)) + s     ; 1-op loop carry on s
```

On a scalar reduction loop this is a measurable steady-state regression (~16% on
an aarch64 ALU-bound microbench). It only shows up when the loop body isn't
vectorized, which is why it has gone unnoticed.

## Fix

Sort expression leaves by an explicit `OpCategory` first, rank second:

| category | placement |
|---|---|
| `LoopCarriedRecurrence` | outermost operand (1-cycle loop carry) |
| `LoopVariant` | by rank, as before |
| `LoopInvariantOrConst` | innermost, so LICM/folding still see them grouped |

With no enclosing loop every leaf is `LoopVariant`, so ordering falls back to
rank-only and matches the previous behavior. Recurrence detection keys on the
phi's parent being the enclosing loop's header with an in-loop incoming edge
(handles multi-latch loops, which `getLoopLatch()` would miss). `LoopInfo` is
built locally in `run()` and skipped for single-BB functions, so it works under
both pass managers without a registered analysis.

## Testing

- New `Transforms/Reassociate/loop-carried-phi.ll` (single/multi-latch, outer
  recurrence with inner sibling, nested-loop LICM preservation, mixed-category,
  fp/fast, self-loop, vector, xor/mul chains, opaque-call-in-body, triple-nested).
- `check-llvm` and `check-llvm-unit` pass; a few `PhaseOrdering` CHECK lines are
  regenerated to reflect the new operand order (no functional change).
- Corpus run over all in-tree `.ll` (~41k files) through `reassociate -verify-each`:
  no crashes or verifier failures.


eapenzhan:
Suggestion: gate the new ordering behind a hidden cl::opt (default on)

This changes the default operand ordering of a very widely-used core pass. A hidden flag defaulting to true would make it easy to:

  - bisect a regression down to this behavior without reverting the commit;
  - A/B the steady-state perf impact (the ~16% is one aarch64 ALU-bound microbench, and reordering reduction chains ripples downstream
  into SLP/scheduling, e.g. the churn in reduce_submuladd.ll);
  - give downstreams a one-line escape hatch.

  It also pairs nicely with the compile-time aspect: if the same flag gates the LoopInfo/DominatorTree construction in run(), turning
  it off removes the per-function DT+LI build cost and drops back to the no-loop classification path — which single_bb_no_loops already
  certifies as legacy-equivalent (with LI == nullptr every non-constant leaf is LoopVariant, so the comparator falls back to
  rank-only; constants still group at the end, same as their rank-0 placement today).

  Matching the existing UseCSELocalOpt style in this file:
```C++
  static cl::opt<bool> ReassociateLoopCarriedRecurrence(
      DEBUG_TYPE "-loop-carried-recurrence",
      cl::desc("Sort loop-carried recurrence phis to the outermost operand of a "
               "reassociated expression, shortening the loop-carried dependency"),
      cl::init(true), cl::Hidden);
```
  and in run():
``` C++
  if (ReassociateLoopCarriedRecurrence && !F.empty() &&
      std::next(F.begin()) != F.end()) {
    DT.emplace(F);
    LIStorage.emplace(*DT);
    LI = &*LIStorage;
  }
```
  This keeps the change on by default while giving a zero-overhead way to turn it off for triage and measurement.

yadongw:
> Suggestion: gate the new ordering behind a hidden cl::opt (default on)
> 
> This changes the default operand ordering of a very widely-used core pass. A hidden flag defaulting to true would make it easy to:
> 
> * bisect a regression down to this behavior without reverting the commit;
> * A/B the steady-state perf impact (the ~16% is one aarch64 ALU-bound microbench, and reordering reduction chains ripples downstream
>   into SLP/scheduling, e.g. the churn in reduce_submuladd.ll);
> * give downstreams a one-line escape hatch.
> 
> It also pairs nicely with the compile-time aspect: if the same flag gates the LoopInfo/DominatorTree construction in run(), turning it off removes the per-function DT+LI build cost and drops back to the no-loop classification path — which single_bb_no_loops already certifies as legacy-equivalent (with LI == nullptr every non-constant leaf is LoopVariant, so the comparator falls back to rank-only; constants still group at the end, same as their rank-0 placement today).
> 
> Matching the existing UseCSELocalOpt style in this file:
> 
> ```c++
>   static cl::opt<bool> ReassociateLoopCarriedRecurrence(
>       DEBUG_TYPE "-loop-carried-recurrence",
>       cl::desc("Sort loop-carried recurrence phis to the outermost operand of a "
>                "reassociated expression, shortening the loop-carried dependency"),
>       cl::init(true), cl::Hidden);
> ```
> 
> and in run():
> 
> ```c++
>   if (ReassociateLoopCarriedRecurrence && !F.empty() &&
>       std::next(F.begin()) != F.end()) {
>     DT.emplace(F);
>     LIStorage.emplace(*DT);
>     LI = &*LIStorage;
>   }
> ```
> 
> This keeps the change on by default while giving a zero-overhead way to turn it off for triage and measurement.

@eapenzhan Good call — added it as -reassociate-loop-carried-recurrence (default on), same style as UseCSELocalOpt. It also gates the DT/LoopInfo build in run(), so flag-off skips that cost and falls back to rank-only ordering. loop-carried-phi-flag.ll covers both states.

llvm/lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -232,6 +242,34 @@ unsigned ReassociatePass::getRank(Value *V) {
...
 }
 
+namespace llvm::reassociate {
+
+OpClassifier::OpClassifier(const LoopInfo *LI, const Instruction *I)
+    : EnclosingLoop(LI && I ? LI->getLoopFor(I->getParent()) : nullptr) {}
+
+OpCategory OpClassifier::classify(const Value *V) const {
+  if (isa<Constant>(V))
+    return OpCategory::LoopInvariantOrConst;
```
taofengliu, line 360: Reassociate also gives all Arguments low ranks.
According to `ReassociatePass::getRank`, maybe we can do it like this:

```suggestion
  if (!isa<Instruction>(V))
    return OpCategory::LoopInvariantOrConst;
  }
   
```
This might have the minimal impact on the original logic.

llvm/lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -2553,6 +2594,20 @@ PreservedAnalyses ReassociatePass::run(Function &F, FunctionAnalysisManager &) {
   // analysing dead basic blocks).
   ReversePostOrderTraversal<Function *> RPOT(&F);
 
+  // Build LoopInfo locally for OpClassifier. We don't take it from the FAM:
+  // Reassociate runs before LICM so it isn't cached yet, and the legacy PM
+  // calls run() with an empty FAM that would assert. Skip single-BB functions
+  // (no loops, so LI stays null and ordering reduces to rank only).
```
taofengliu: FAM.getResult<>() is lazy and will compute on demand. Using FAM has a concrete benefit: the computed DT/LI gets cached, so subsequent passes (LICM runs right after Reassociate in both pipeline positions) can reuse them instead of rebuilding from scratch.
For the legacy PM, the DummyFAM issue can be addressed by registering the required analyses in the legacy wrapper

taofengliu (commented):
First of all. In line 1507 and 1512, there are unclassified `ValueEntry`s.

And I found two scenarios that are not handled correctly.

## False-positive header phi classification can worsen loop carry
For IR:
```
define i32 @false_positive_induction_steals_outermost(i32 %n, ptr %arr) {
entry:
  br label %loop

loop:
  %s = phi i32 [ 0, %entry ], [ %result, %loop ]
  %i = phi i32 [ 0, %entry ], [ %i.next, %loop ]
  %p = getelementptr i32, ptr %arr, i32 %i
  %v = load i32, ptr %p, align 4
  %body = mul i32 %v, 7
  %t1 = add i32 %s, %i
  %result = add i32 %t1, %body
  %i.next = add i32 %i, 1
  %cmp = icmp slt i32 %i.next, %n
  br i1 %cmp, label %loop, label %exit

exit:
  ret i32 %result
}
```
After Reassociate, we get this:
```
define i32 @false_positive_induction_steals_outermost(i32 %n, ptr %arr) {
entry:
  br label %loop

loop:                                             ; preds = %loop, %entry
  %s = phi i32 [ 0, %entry ], [ %result, %loop ]
  %i = phi i32 [ 0, %entry ], [ %i.next, %loop ]
  %p = getelementptr i32, ptr %arr, i32 %i
  %v = load i32, ptr %p, align 4
  %body = mul i32 %v, 7
  %t1 = add i32 %s, %body
  %result = add i32 %t1, %i
  %i.next = add i32 %i, 1
  %cmp = icmp slt i32 %i.next, %n
  br i1 %cmp, label %loop, label %exit

exit:                                             ; preds = %loop
  ret i32 %result
}
```
The `%s` should be the outermost, but not in the current implementation.
Expected IR after Reassociate:
```
define i32 @false_positive_induction_steals_outermost(i32 %n, ptr %arr) {
entry:
  br label %loop

loop:                                             ; preds = %loop, %entry
  %s = phi i32 [ 0, %entry ], [ %result, %loop ]
  %i = phi i32 [ 0, %entry ], [ %i.next, %loop ]
  %p = getelementptr i32, ptr %arr, i32 %i
  %v = load i32, ptr %p, align 4
  %body = mul i32 %v, 7
  %t1 = add i32 %i, %body
  %result = add i32 %t1, %s
  %i.next = add i32 %i, 1
  %cmp = icmp slt i32 %i.next, %n
  br i1 %cmp, label %loop, label %exit

exit:                                             ; preds = %loop
  ret i32 %result
}
```

## CSE pair reordering can override OpCategory sort
For IR:
```
declare void @use(i32)

define i32 @cse_overrides_recurrence(i32 %n, ptr %arr) {

entry:
  br label %loop

loop:
  %s = phi i32 [ 0, %entry ], [ %result, %loop ]
  %i = phi i32 [ 0, %entry ], [ %i.next, %loop ]
  %p = getelementptr i32, ptr %arr, i32 %i
  %v = load i32, ptr %p, align 4
  %a = mul i32 %v, 3
  %b = mul i32 %v, 5
  %b2 = lshr i32 %v, 2

  %side1 = add i32 %s, %a
  %side  = add i32 %side1, %b
  call void @use(i32 %side)

  %t1 = add i32 %s, %a
  %result = add i32 %t1, %b2

  %i.next = add i32 %i, 1
  %cmp = icmp slt i32 %i.next, %n
  br i1 %cmp, label %loop, label %exit

exit:
  ret i32 %result
}
```
After Reassociate, we get this:
```
declare void @use(i32)

define i32 @cse_overrides_recurrence(i32 %n, ptr %arr) {
entry:
  br label %loop

loop:                                             ; preds = %loop, %entry
  %s = phi i32 [ 0, %entry ], [ %result, %loop ]
  %i = phi i32 [ 0, %entry ], [ %i.next, %loop ]
  %p = getelementptr i32, ptr %arr, i32 %i
  %v = load i32, ptr %p, align 4
  %a = mul i32 %v, 3
  %b = mul i32 %v, 5
  %b2 = lshr i32 %v, 2
  %side1 = add i32 %s, %a
  %side = add i32 %side1, %b
  call void @use(i32 %side)
  %t1 = add i32 %s, %a
  %result = add i32 %t1, %b2
  %i.next = add i32 %i, 1
  %cmp = icmp slt i32 %i.next, %n
  br i1 %cmp, label %loop, label %exit

exit:                                             ; preds = %loop
  ret i32 %result
}
```
We can see that `%s` is not put in the outermost position in the recurrence chain.
We should prioritize evaluating the CSE and recurrence chain length.

taofengliu:
Do we have any data on how this PR affects actual performance?

yadongw:
@taofengliu thanks, all fixed.

- **xor 1507/1512:** classified. Consistency only — `OptimizeXor` doesn't re-sort, so output is unchanged; I didn't add a test that wouldn't discriminate. Got a case where it should? I'll add one.
- **Induction phi:** the recurrence is now the header phi whose in-loop incoming value is this expression's root, so `%i` stays loop-variant and `%s` keeps the outermost slot. Test `induction_phi_not_recurrence`.
- **CSE pair:** the pair-candidate loop skips any pair containing a recurrence operand. Test `cse_pair_keeps_recurrence_outermost`.

Both tests fail without the fix.

llvm/lib/Transforms/Scalar/Reassociate.cpp
yadongw: Kept `isa<Constant>`. Inside loops `isLoopInvariant` already buckets arguments as invariant, so this only differs outside loops — and there it splits `X` from `~X`, which `getRank` deliberately keeps at the same rank (so they sort adjacent and `X & ~X → 0`, `X | ~X → -1`, `X + -X → 0` fire). Bucketing the arg as invariant breaks that adjacency: `inverses.ll` `@test1` regresses `ret i32 0` → un-folded `and` chain (6/89 fail). Happy to take it if you have a case where it actually helps.

llvm/lib/Transforms/Scalar/Reassociate.cpp
yadongw: Adopted — `LI` comes from `FAM.getResult<LoopAnalysis>` (Reassociate preserves the CFG, so the LICM that runs right after reuses it); the legacy wrapper supplies it via `LoopInfoWrapperPass` (`INITIALIZE_PASS_DEPENDENCY` + a flag-gated `addRequired`), shared body in `runImpl`. Single-BB functions skip `LI` and fall back to rank-only. (The legacy PM isn't lit-testable — `opt` dropped `-enable-new-pm=0` — so the new-PM tests cover `runImpl`.)

yadongw:
@taofengliu yes. Flag on vs off on the same build, median ns/element:

- `s += v*7 + (v>>2)`: 0.687 → 0.459 (+33%)
- 4 body ops: +23%; 3 body ops: +5%
- single-op / non-add / no-fast-math: within noise, no regressions

The gain scales with the loop-carried add-chain length. On the SLP ripple — the `slpordering.ll` / `reduce_submuladd.ll` churn is a reschedule, not a regression: `opt -O3` flag on vs off gives identical vector op counts (37 `shufflevector`, 6 `add` + 6 `sub`, 2 `vector.reduce`). Full `llvm/test/Transforms` is clean (the one failure is an unrelated `ThinLTOBitcodeWriter` test that runs no reassociate).

llvm/lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -232,6 +241,37 @@ unsigned ReassociatePass::getRank(Value *V) {
...
+  // in-loop incoming value is the root we are reassociating. Other header phis
+  // (e.g. induction variables, whose backedge value is a different chain) are
+  // ordinary loop-variant leaves and must not steal the outermost slot.
+  // contains() rather than getLoopLatch(), which is null for multi-latch loops.
+  if (const auto *Phi = dyn_cast<PHINode>(V))
+    if (Phi->getParent() == EnclosingLoop->getHeader())
+      for (BasicBlock *Pred : Phi->blocks())
+        if (EnclosingLoop->contains(Pred) &&
+            Phi->getIncomingValueForBlock(Pred) == RootI)
+          return OpCategory::LoopCarriedRecurrence;
```
taofengliu: This is too strict for LoopCarriedRecurrence. Any intermediate operation between the expression result and the phi's backedge value breaks the classification entirely, causing the recurrence phi to be treated as LoopVariant and sink to the innermost position by rank.

For a simple example:
```
define i32 @indirect_recurrence_and_clamp(i32 %n, ptr %arr) {
entry:
  br label %loop

loop:
  %s = phi i32 [ 0, %entry ], [ %clamped, %loop ]
  %i = phi i32 [ 0, %entry ], [ %i.next, %loop ]
  %p = getelementptr i32, ptr %arr, i32 %i
  %v = load i32, ptr %p, align 4
  %m = mul i32 %v, 7
  %r = lshr i32 %v, 2
  %add1 = add i32 %m, %r
  %add2 = add i32 %add1, %s
  %clamped = and i32 %add2, 255
  %i.next = add i32 %i, 1
  %cmp = icmp slt i32 %i.next, %n
  br i1 %cmp, label %loop, label %exit

exit:
  ret i32 %clamped
}
```


This is just a simple example to illustrate the fragility of the current handling. The detection of LoopCarriedRecurrence needs to consider more possible LLVM IR patterns.

taofengliu (commented):
This PR will have a relatively broad impact on LLVM optimizations as a whole, since the reassociate pass affects other optimizations. I think we need more comprehensive performance regression testing. 

Although I tested it via https://github.com/llvm/llvm-test-suite.git and produced fairly reliable data, I am still not entirely confident about its overall impact on LLVM optimizations. We can land it once the left improvements are in place with performance data we expected. A more prudent approach would be to submit this PR upstream to get feedback from the LLVM community reviewers, which doesn't block the current PR from being merged.

Additionally, since this PR is a generic optimization independent of PGO, I suggest we land the PGO patches separately.

taofengliu:
Here are the tests result with https://github.com/llvm/llvm-test-suite.git

## Results Overview

| Suite | Total | Reliable | Improved (>2%) | Regressed (>2%) | Neutral | Geomean Speedup | Avg % Change |
|-------|-------|----------|----------------|-----------------|---------|-----------------|--------------|
| LoopVectorization | 696 | 696 | 12 | 20 | 664 | 0.9982x | -0.17% |
| MultiSource | 167 | 149 | 15 | 16 | 118 | 1.0006x | -0.16% |

## LoopVectorization Detailed Analysis

### Distribution of Changes

| Range | Count |
|-------|-------|
| < -10% | 3 |
| -10% to -5% | 6 |
| -5% to -2% | 11 |
| -2% to +2% | 664 |
| +2% to +5% | 8 |
| +5% to +10% | 3 |
| > +10% | 1 |

### Improvements (Median > 2%)

| Benchmark | Median % | Mean % | PR Median (us) | Upstream Median (us) | PR CV% | Upstream CV% |
|-----------|----------|--------|----------------|---------------------|--------|-------------|
| IC1VW4LoopTC31 | +13.07% | +4.48% | 8.12 | 9.18 | 7.3 | 7.2 |
| IC1VW4LoopTC7 | +6.97% | +2.03% | 3.81 | 4.08 | 6.1 | 4.2 |
| AutoVecForBigLoopTC32 | +5.40% | +5.17% | 5.34 | 5.63 | 2.8 | 1.4 |
| autovec_no_early_exit_single_load\<uint8_t\>/999 | +5.21% | +4.30% | 51.47 | 54.15 | 0.3 | 1.3 |
| autovec_early_exit_taken_mid_single_load\<uint32_t\>/10 | +4.20% | +1.59% | 5.87 | 6.12 | 6.2 | 6.2 |
| IC1VW4LoopWithReductionTC31 | +4.17% | +1.02% | 5.69 | 5.93 | 9.4 | 8.4 |
| LoopWithReductionAutoVecTC31 | +4.05% | -0.71% | 6.33 | 6.59 | 6.4 | 11.2 |
| AutoVecForLoopTC31 | +3.97% | +3.89% | 6.27 | 6.52 | 4.0 | 0.1 |
| AutoVecForBigLoopTC8 | +3.88% | +2.26% | 2.55 | 2.65 | 2.7 | 0.1 |
| IC2VW4LoopWithReductionTC31 | +3.32% | +1.26% | 6.49 | 6.70 | 2.0 | 1.8 |
| benchReductionAutoVec\<uint8_t\>/127 | +2.68% | +2.77% | 25.11 | 25.78 | 0.0 | 0.6 |
| benchAutoVec\<uint32_t\>/65 | +2.47% | +1.50% | 10.07 | 10.32 | 1.7 | 1.9 |

### Regressions (Median < -2%)

| Benchmark | Median % | Mean % | PR Median (us) | Upstream Median (us) | PR CV% | Upstream CV% |
|-----------|----------|--------|----------------|---------------------|--------|-------------|
| LoopWithReductionAutoVecTC7 | -21.65% | -14.50% | 3.73 | 2.93 | 2.7 | 13.9 |
| IC2VW4LoopTC7 | -21.02% | -7.58% | 5.56 | 4.39 | 13.1 | 14.1 |
| IC2VW4LoopTC31 | -11.60% | -8.27% | 7.01 | 6.19 | 7.2 | 0.1 |
| IC2VW4LoopWithReductionTC63 | -6.88% | -6.67% | 9.02 | 8.40 | 2.9 | 3.7 |
| AutoVecForBigLoopTC15 | -6.73% | -4.91% | 4.09 | 3.82 | 4.1 | 0.5 |
| IC1VW4LoopWithReductionTC15 | -6.46% | -4.89% | 4.09 | 3.82 | 2.6 | 2.7 |
| multi_csa_only_autovec_int64 | -6.43% | -6.41% | 73908.6 | 69155.8 | 5.0 | 2.3 |
| multi_csa_only_novec_int64 | -6.25% | -6.28% | 73923.9 | 69303.3 | 5.1 | 2.1 |
| benchForTruncOrZextVecInLoopVW8 u8->u16 | -5.20% | -5.11% | 294.38 | 279.06 | 0.3 | 0.0 |
| csa_with_in_loop_arith_autovec_int64 | -4.87% | -4.82% | 76845.2 | 73103.5 | 3.7 | 1.3 |
| AutoVecForLoopTC15 | -4.18% | -2.64% | 3.94 | 3.78 | 1.6 | 3.8 |
| csa_with_in_loop_arith_novec_int64 | -4.13% | -4.54% | 76541.2 | 73380.8 | 3.7 | 1.2 |
| IC4VW4BigLoopTC7 | -3.62% | -1.75% | 18.02 | 17.37 | 2.2 | 1.6 |
| IC4VW4BigLoopTC8 | -3.58% | -1.68% | 20.17 | 19.45 | 2.5 | 1.6 |
| benchForTruncOrZextVecInLoop u8->u16 | -2.98% | -3.00% | 296.68 | 287.83 | 0.2 | 0.1 |
| IC2VW4BigLoopTC7 | -2.95% | -1.25% | 18.13 | 17.59 | 1.7 | 1.4 |
| autovec_early_exit_taken_mid u16/256 | -2.93% | -4.90% | 18.76 | 18.22 | 6.2 | 0.1 |
| IC4VW4BigLoopTC15 | -2.60% | -0.37% | 35.97 | 35.04 | 1.4 | 2.5 |
| IC2VW1LoopWithReductionTC7 | -2.44% | -0.88% | 3.81 | 3.72 | 3.0 | 2.9 |
| benchForTruncOrZextVecInLoopVW16 u8->u16 | -2.20% | -2.12% | 251.34 | 245.81 | 0.1 | 0.2 |

## MultiSource Detailed Analysis

### Distribution of Changes (149 Reliable Benchmarks)

| Range | Count |
|-------|-------|
| < -10% | 1 |
| -10% to -5% | 4 |
| -5% to -2% | 11 |
| -2% to +2% | 118 |
| +2% to +5% | 9 |
| +5% to +10% | 6 |
| > +10% | 0 |

### Improvements (Median > 2%)

| Benchmark | Median % | Mean % | PR Median (ms) | Upstream Median (ms) | PR CV% | Upstream CV% | Trustworthiness |
|-----------|----------|--------|----------------|---------------------|--------|-------------|-----------------|
| kimwitu++/kc | +7.91% | -3.57% | 0.01 | 0.01 | 24.5 | 28.7 | Low (high CV, sub-ms) |
| uudecode | +7.69% | +5.41% | 0.01 | 0.01 | 0.9 | 5.2 | Low (sub-ms) |
| Olden/mst | +7.21% | +6.03% | 0.02 | 0.02 | 1.6 | 6.1 | Low (sub-ms) |
| Olden/health | +6.17% | +0.43% | 0.10 | 0.11 | 7.2 | 5.1 | Moderate |
| beamformer | +5.21% | +4.97% | 0.47 | 0.50 | 0.7 | 5.3 | Moderate (sub-ms but larger) |
| automotive-susan | +5.11% | -8.68% | 0.02 | 0.02 | 20.5 | 3.5 | Low (high CV, mean/median disagree) |
| Olden/voronoi | +4.40% | +2.75% | 0.13 | 0.13 | 4.9 | 1.8 | Moderate |
| network-patricia | +4.06% | +1.58% | 0.04 | 0.05 | 1.4 | 5.7 | Low (sub-ms) |
| McCat/bintr | +3.95% | +2.39% | 0.07 | 0.07 | 0.7 | 2.6 | Low (sub-ms) |
| HPCCG | +3.83% | +0.83% | 0.42 | 0.44 | 4.6 | 6.0 | Moderate |
| McCat/imp | +3.50% | +4.97% | 0.03 | 0.03 | 1.9 | 1.6 | Low (sub-ms) |
| gnugo | +3.35% | +0.79% | 0.02 | 0.02 | 2.6 | 2.6 | Low (sub-ms) |
| telecomm-CRC32 | +3.24% | +1.63% | 0.06 | 0.06 | 1.5 | 4.4 | Low (sub-ms) |
| ClamAV/clamscan | +2.44% | +1.51% | 0.07 | 0.07 | 2.8 | 2.3 | Low (sub-ms) |
| siod | +2.39% | +2.08% | 0.88 | 0.90 | 0.5 | 0.4 | High |

### Regressions (Median < -2%)

| Benchmark | Median % | Mean % | PR Median (ms) | Upstream Median (ms) | PR CV% | Upstream CV% | Trustworthiness |
|-----------|----------|--------|----------------|---------------------|--------|-------------|-----------------|
| ALAC/decode | -28.33% | -29.48% | 0.01 | 0.01 | 18.6 | 1.4 | Low (high CV, sub-ms) |
| MallocBench/gs | -7.65% | -5.33% | 0.02 | 0.02 | 1.0 | 3.2 | Low (sub-ms) |
| make_dparser | -6.67% | -3.19% | 0.01 | 0.01 | 5.8 | 6.0 | Low (sub-ms) |
| JM/ldecod | -5.67% | -3.90% | 0.02 | 0.02 | 2.8 | 3.6 | Low (sub-ms) |
| telecomm-FFT | -5.15% | -10.72% | 0.02 | 0.02 | 13.6 | 8.7 | Low (high CV, sub-ms, mean/median disagree) |
| minisat | -4.93% | -4.06% | 3.52 | 3.35 | 3.3 | 0.4 | **High** |
| ALAC/encode | -4.84% | -11.76% | 0.02 | 0.02 | 19.2 | 3.0 | Low (high CV, sub-ms, mean/median disagree) |
| Olden/tsp | -4.63% | -4.83% | 0.31 | 0.29 | 1.3 | 1.5 | Moderate |
| telecomm-gsm | -4.55% | -19.83% | 0.07 | 0.07 | 29.9 | 0.9 | Low (very high CV, mean/median disagree) |
| llubenchmark | -3.25% | -17.20% | 6.31 | 6.11 | 26.6 | 2.5 | Low (very high PR CV) |
| TSVC/Expansion-dbl | -2.80% | -1.64% | 1.46 | 1.42 | 1.8 | 1.3 | Moderate |
| consumer-typeset | -2.77% | +0.05% | 0.07 | 0.06 | 3.3 | 4.2 | Low (sub-ms, mean/median disagree) |
| SPASS | -2.70% | -9.05% | 4.58 | 4.46 | 13.1 | 0.5 | Low (high PR CV) |
| automotive-basicmath | -2.34% | -1.74% | 0.15 | 0.15 | 2.3 | 1.4 | Low (sub-ms) |
| fhourstones3.1 | -2.12% | +0.88% | 0.64 | 0.62 | 7.9 | 10.6 | Low (median/mean disagree) |
| Olden/bisort | -2.06% | -2.59% | 0.27 | 0.27 | 1.9 | 2.0 | Low (sub-ms) |

llvm/lib/Transforms/Scalar/Reassociate.cpp
yadongw: Fixed. Replaced the exact `phi.backedge == RootI` test with a bounded backward
walk (`rootReaches`) from the backedge value through in-loop operands, stopping at
phis — so a recurrence closing through intermediate ops (clamp/mask/...) is
recognised. Stopping at phis keeps induction variables out.

New discriminating tests: `indirect_recurrence_and_clamp`, `recurrence_through_two_ops`,
`two_recurrences_one_expression`, `cse_pair_indirect_recurrence`,
`backedge_through_phi_not_recurrence`.

yadongw:
@taofengliu I built and ran the flagged benchmarks end-to-end, flag on vs off
(interleaved, pinned to one core) — none reproduces:

| benchmark | reported | off → on |
|---|---|---|
| minisat (short.cnf) | -4.93% | 1.813 → 1.806 s |
| Olden/tsp (1024000) | -4.63% | 0.1677 → 0.1677 s |
| TSVC/Expansion-dbl | -2.80% | 0.873 → 0.879 s |
| LV csa int64 | -6.4% | within noise (same-binary spread 2.67% > A/B 0.24%) |

All within run-to-run noise. The asm for each is byte-identical on/off or differs
only by commutative operand order — none has a 3+-leaf reassociable recurrence,
the only shape this touches (there it shortens the carry: 1357 → 656 ns/call,
2.07x). Frontend clang-16, codegen the PR build, aarch64 + x86.

yadongw:
Coverage and perf update.

Tests (each fails without the fix): closure that is direct / through one op (mask)
/ through a two-op chain; two recurrences in one chain; the CSE-pair skip with an
indirectly-closed recurrence; plus negatives — a phi reaching the root only via
another phi stays untagged (phi-stop), and an induction variable stays untagged.
Full Reassociate lit suite green on aarch64 and x86.

Perf (recurrence ordering on vs off, loop-carried reductions, end-to-end through
the JIT, ns/elem): `s += v*7 + (v>>2)` 0.68 → 0.45 (-34%); 4-op chain 0.69 → 0.53
(-24%); 3-op -4.5%. Kernels with no multi-op recurrence chain (dot, hash, xor,
product, two-acc, strict-FP) are unchanged (±0.2%). aarch64 measured; x86 codegen
is identical-or-commutative and the PGO jtreg suite is green there too.

llvm/lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -232,6 +241,68 @@ unsigned ReassociatePass::getRank(Value *V) {
...
+  // reach the root and stay ordinary loop-variant leaves, so they don't steal
+  // the outermost slot. contains() rather than getLoopLatch(), which is null
+  // for multi-latch loops.
+  if (const auto *Phi = dyn_cast<PHINode>(V))
+    if (Phi->getParent() == EnclosingLoop->getHeader())
+      for (BasicBlock *Pred : Phi->blocks())
+        if (EnclosingLoop->contains(Pred) &&
+            rootReaches(Phi->getIncomingValueForBlock(Pred), RootI,
+                        EnclosingLoop))
+          return OpCategory::LoopCarriedRecurrence;
```
taofengliu: We may still be missing some common cases of "LoopCarriedRecurrence", which may affect the effectiveness of the optimization.

For example:

(1) Loop carried recurrence is not a phi node
```
loop:
  %s       = phi i32 [ 0, %entry ], [ %root, %loop ]
  %sscaled = shl i32 %s, 1          ; ← loop-carried
  %m  = mul i32 %v, 7
  %r  = lshr i32 %v, 2
  %add1 = add i32 %m, %r
  %root = add i32 %add1, %sscaled
```

(2) `rootReaches` skips all phi nodes
```
loop:
  %s = phi i32 [ 0, %entry ], [ %merged, %latch ]
  %add1 = add i32 %m, %r
  %root = add i32 %add1, %s          ; RootI
  br i1 %cond, label %then, label %else

then:  %t = add i32 %root, 1 ; br label %latch

else:                              br label %latch

latch:
  %merged = phi i32 [ %t, %then ], [ %root, %else ]   ; ← skipped phi
```

I had the AI derive the complete condition for LoopCarriedRecurrence we can refer to. (If we want to find out as many LoopCarriedRecurrences as possible.)
```
Define R = intra-iteration, in-loop, forward def-use reachability. Edge X→Y exists iff X is an operand of Y and both are in L, except an edge that enters a phi through a back-edge incoming is forbidden (that incoming is the carry from the previous iteration, not this one). A phi-incoming (V_in, InBB → PhiBB) is a back-edge/carry iff PhiBB is a loop header and InBB is inside that loop — detectable purely from LoopInfo (getLoopFor(PhiBB)->getHeader()==PhiBB && contains(InBB)). Consequence drawn directly from this:
  - Header phi latch-incomings are carries → blocked (they close the previous iteration).
  - Non-header (merge) phi incomings are all intra-iteration → crossable (then/else/latch merges happen within one iteration).
  - Non-phi ops always crossable (shl, and, select, add, …).
  
Exact theorem. V is a LoopCarriedRecurrence ⟺ V ∈ forwardR(P) for some reduction phi P that RootI closes, where:
  - P is a reduction phi = a header phi of L with a latch incoming (the true accumulator; merge-phis and induction vars that aren't on this cycle are excluded).
  - "RootI closes P" = RootI R-reaches V_close(P) (the value P takes from its latch) — i.e. RootI is on the intra-iteration compute path that produces P_{n+1}.
  - forwardR(P) = the intra-iteration forward slice from P (P itself included) = "values whose value this iteration depends on P_n".
```

And suggested implementation:
```
Per expression: (1) compute SeenFromRoot = forwardR(RootI); (2) closed reductions = {P : V_close(P) ∈ SeenFromRoot}; (3) if none, no leaf is a recurrence (RootI isn't on any carry → today's fallback); else Late = ⋃ forwardR(P) (multi-source); (4) classify(V) = V ∈ Late. Exact, O(loop body) per expression, O(1) per leaf, and it replaces the phi-only-leaf test + the 32-step rootReaches with one uniform mechanism. Constructor signature (LI, RootI) is unchanged.
```

llvm/lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -77,6 +78,14 @@ static cl::opt<bool>
                             "when exposing CSE opportunities"),
                    cl::init(true), cl::Hidden);
 
+// When off, run() skips the LoopInfo fetch and OpClassifier sees a null loop,
+// so ordering falls back to rank-only (legacy behavior).
+static cl::opt<bool> ReassociateLoopCarriedRecurrence(
```
taofengliu, line 83: I didn't catch this earlier.

`ReassociateLoopCarriedRecurrence=false` is not strictly equivalent to pre-PR behavior.

With the flag off, run() skips the LoopInfo fetch, so OpClassifier sees `EnclosingLoop == nullptr` and `classify()` degenerates to: non-constants → LoopVariant (1), constants → LoopInvariantOrConst (2). The comparison is category-first, rank-second, whereas  pre-PR is purely `return LHS.Rank > RHS.Rank;`. These two agree only under the invariant "every constant has rank 0 and no non-constant has rank 0."  But that invariant is not guaranteed.

Suggested fix (make the category dimension a no-op whenever there's no enclosing loop, so the comparator provably degenerates to rank-only):

```
  // Reassociate.cpp — classify() (278)
  OpCategory OpClassifier::classify(const Value *V) const {
    if (!EnclosingLoop)                        // move BEFORE the isa<Constant> check
      return OpCategory::LoopVariant;          // incl. constants → uniform → rank-only
    if (isa<Constant>(V))
      return OpCategory::LoopInvariantOrConst;
    // ...
  }

  // Reassociate.cpp — OptimizeExpression() (2003)
  Ops.push_back(ValueEntry(0, Cst, Cls.classify(Cst)));   // don't bypass the classifier
```

(Alternative, heavier but maximally robust: gate operator< on the flag itself so it returns LHS.Rank > RHS.Rank when off.)

llvm/lib/Transforms/Scalar/Reassociate.cpp
yadongw: 改成前向 def-use 切片(`forwardLoopSlice`):从 root 找到它闭合的 reduction phi,再前向切片得 recurrence 集。非 phi 携带叶、merge phi 闭合、内层累加器都覆盖了,加了测试。

另外只对带 recurrence 的表达式生效,普通循环表达式保持原 rank;`debugloc-reass-add.ll` 已还原。

llvm/lib/Transforms/Scalar/Reassociate.cpp
yadongw: 已修。flag-off 时所有叶归一类、退化为纯 rank,`OptimizeExpression` 的折叠常量也走 classify;开关字节一致。

PR #74, opened 2026-06-12
feat: constant folding

### LINKED ISSUE #509: Const oops' types are not handle by JavaType interface.
**Goals**
Make JavaType interface aware of const oops' types.

**Description**
JavaType interface is used to query a value's type for LLVM passes, but it only read types from attributes and metadatas. Const oops are declared as global variables, and load instructions for them are not attached with metadata

taofengliu:
Implement constant folding for static final fields.
Extend VMCallbackLog to support string arg/result
Also fix https://github.com/jeandle/jeandle-jdk/issues/509

TODO:
- Support stable fields
- Support const array elements folding


llvm/lib/IR/Jeandle/VMCallbackLog.cpp
```diff
@@ -283,13 +336,57 @@ static std::optional<int64_t> parseValue(StringRef Token,
       return std::nullopt;
     return static_cast<int64_t>(Val);
   }
+  case VMCallbackValueType::Long: {
+    int64_t val;
```
Cruise20:
```suggestion
    int64_t Val;
```

llvm/include/llvm/IR/Jeandle/VMCallbackLog.h
```diff
@@ -75,6 +76,38 @@ enum CallbackKind : unsigned { ALL_JEANDLE_VM_CALLBACKS(DEF_CALLBACK_KIND) };
...
+    if (IsString != O.IsString)
+      return IsString < O.IsString;
+    return IsString ? (StrVal < O.StrVal) : (NumVal < O.NumVal);
+  }
+};
+
+} // namespace llvm::jeandle
+
+namespace llvm::jeandle {
+
```
Cruise20:
```suggestion
```

Cruise20 (approved):

PR #75, opened 2026-06-15
feat: Add eliminate safepoint pass

eapenzhan:
## Related issue(s):

issue #70 

The frontend emits a jeandle.safepoint_poll at every bytecode backward branch and method return. Each is an opaque call carrying deopt state, so it blocks LLVM loop optimizations (vectorization, LICM, unrolling) and costs a TLS load + branch every iteration. Many of these polls are redundant once the IR is built — duplicated across CFG merges, or sitting in loops short enough that no poll is needed at all.

  Solution. A new SafepointElimination function pass that removes only provably-safe polls, keeping every loop within a bounded time-to-safepoint:

  - Adjacent collapse — back-to-back polls on a straight-line path collapse to the later one (mirrors C2 SafePointNode::Identity); runs function-wide, so inlining-produced duplicates outside loops are caught too.
  - Keep-one dedup — when a poll dominates a loop's latch, the loop's other polls are redundant; keep the latch-closest one (C2 remove_safepoints(keep_one)). With no dominating poll, nothing is deleted.
  - Short-loop deletion — an innermost loop whose SCEV max backedge-taken count provably fits -jeandle-safepoint-chunk-iters (default 1000) drops its polls entirely.

  Safety invariants: a poll is deleted only when another poll keeps covering the loop or the trip count is provably small; side-exit/deopt polls are never touched; the pass bails on irreducible CFGs and on non-Java / template-helper functions; polls are only ever deleted or relocated, never synthesized. A default-off verify<jeandle-safepoint-coverage> pass asserts the invariant, and utils/jeandle/count-safepoints.py classifies polls vs. lowered statepoints for triage. Everything is behind the-jeandle-enable-safepoint-elim master switch.

  This is the foundation only. Strip mining for long/unknown-bound loops, a second cleanup pass after the default pipeline, and call-dominates-latch elimination are follow-up PRs.


llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
```diff
@@ -0,0 +1,295 @@
...
+// they don't break the back-to-back adjacency of two polls.
+bool isAdjacencyTransparent(const Instruction &I) {
+  return I.isDebugOrPseudoInst() || I.isLifetimeStartOrEnd();
+}
+
+// Collapse runs of adjacent polls within a block, keeping the last one: the
+// earlier poll's deopt state belongs to an upstream program point that the
+// later poll's state supersedes (the converse — reusing an earlier state for
+// a later point — would be wrong). A coverage-marked poll survives over an
+// unmarked one regardless of position.
```
taofengliu: The comment here might be slightly misleading — the deopt state carried by each safepoint poll is self-consistent on its own; there is no need to conflate/mix the deopt states across different safepoint polls.

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
```diff
@@ -0,0 +1,295 @@
...
+bool collapseAdjacentPolls(BasicBlock &BB) {
```
taofengliu: The current implementation only has visibility into safepoint polls within a single block. If multiple blocks are connected via unconditional branches, would it be possible to merge/handle them as a chain together?

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
```diff
@@ -0,0 +1,295 @@
...
+  } else {
+    auto &DT = AM.getResult<DominatorTreeAnalysis>(F);
+    auto &SE = AM.getResult<ScalarEvolutionAnalysis>(F);
+
+    // Innermost-first, so an inner loop settles its own coverage before any
+    // outer-loop decision looks at the nest.
+    SmallVector<Loop *, 8> Loops(LI.getLoopsInPreorder());
+    for (Loop *L : llvm::reverse(Loops)) {
+      Changed |= keepOneLoopPoll(*L, LI, DT);
+      Changed |= deleteShortLoopPolls(*L, SE);
```
taofengliu: keepOneLoopPoll restricts its coverage search to blocks the loop owns directly (LI.getLoopFor(BB) != &L, SafepointElimination.cpp:151/166), so it never recognizes that a sub-loop's poll can cover the enclosing
  loop. When a sub-loop poll dominates the outer latch, every outer iteration already passes it, and all of the outer's own polls are redundant — yet the transform keeps one.

And also, the comment's justification (SafepointElimination.cpp:140-143) — that inner polls "may legitimately disappear later under the inner loop's own trip-count proof" — doesn't hold against the actual processing order. run() (:281-285) processes loops innermost-first (reverse(getLoopsInPreorder())), and each loop gets keepOneLoopPoll + deleteShortLoopPolls before its parent is touched. By the time the outer loop is processed, the inner's polls are already finalized

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
taofengliu: If it's hard to implement, the current method is OK. We can discuss and do it in the future.

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
eapenzhan: Good catch — the "supersedes / reusing an earlier state" wording was misleading; nothing is merged or moved, each poll's deopt state is self-contained, so dropping either adjacent poll is safe. Reworded: the two are simply redundant, we keep the later by convention (as SafePointNode::Identity does) plus the coverage-marked one if marked.

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
eapenzhan: Possible, but largely covered already in practice: SimplifyCFG merges single-pred/single-succ unconditional chains into one block, after which the block-local sweep catches them — and the second SafepointElimination run (after the default pipeline, which includes SimplifyCFG) sees that merged form. Cross-block redundancy inside loops is separately handled by keep-one via dominance. The only uncovered case is an unmerged non-loop unconditional chain, which I haven't seen in real IR, so I'd prefer to defer cross-block chaining to keep the peephole simple — happy to add it if you have a concrete case.

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
eapenzhan: You're right on both. (1) The comment's reasoning is wrong — loops are processed innermost-first, so a sub-loop's polls are finalized before the parent is processed; they don't "disappear later under the inner trip-count proof." Fixed the comment. (2) We do leave the sub-loop-dominating-poll dedup on the table, intentionally: if this loop dropped its own polls to lean on a sub-loop's poll, a later transform (full unroll / LoopDeletion / vectorize / second-pass run) that removes that poll would leave this loop naked, and keep-one can't foresee it — so we keep each loop self-covered (also makes the invariant locally verifiable). Capturing the optimization would require marking the sub-loop poll as also covering the outer loop so downstream/short-delete won't drop it; I'd suggest a follow-up rather than this PR.

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
```diff
@@ -0,0 +1,1345 @@
...
+        Victim->eraseFromParent();
+        Changed = true;
+        Prev = Survivor;
+        continue;
+      }
+      Prev = Cur;
+      continue;
+    }
+    if (!isAdjacencyTransparent(I))
+      Prev = nullptr;
```
taofengliu: Can we treat all non-CallBase instructions as `isAdjacencyTransparent`?

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
```diff
@@ -0,0 +1,1345 @@
...
+// would leave the loop naked if a later transform (full unroll, LoopDeletion,
+// vectorization, the second-pass run) removes that sub-loop poll. Keeping each
+// loop self-covered also keeps the coverage invariant locally verifiable.
+// (Loops are processed innermost-first, so a sub-loop's polls are already
+// finalized here — the concern is downstream passes, not this one.)
+bool keepOneLoopPoll(Loop &L, LoopInfo &LI, DominatorTree &DT,
+                     ScalarEvolution &SE) {
+  BasicBlock *Latch = L.getLoopLatch();
+  if (!Latch)
+    return false;
```
taofengliu: Do we need to run `LoopSimplifyPass` before to make there is only one latch for every loop? Else there maybe more than one back edges for a loop.

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
```diff
@@ -0,0 +1,1345 @@
...
+// nothing observable between them are redundant; each carries its own
+// self-contained deopt state, so dropping either is safe (no state is merged
+// or moved). We keep the later by convention (as C2's SafePointNode::Identity
+// does), and the coverage-marked one if either is marked, regardless of
+// position.
+bool collapseAdjacentPolls(BasicBlock &BB) {
+  bool Changed = false;
+  CallInst *Prev = nullptr;
+  for (Instruction &I : llvm::make_early_inc_range(BB)) {
+    if (isSafepointPoll(I)) {
```
taofengliu: Here we can assume a Java call as a safepoint poll

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
```diff
@@ -0,0 +1,1851 @@
...
+// loop's other polls are redundant: keep the latch-closest dominating poll (a
+// coverage-marked one wins) and erase the rest. Without a dominating poll,
+// deleting any poll could leave an iteration path uncovered — delete nothing.
+//
+// Only polls in blocks the loop owns directly are considered; a sub-loop's
+// poll is not taken as this loop's coverage even if it dominates the latch.
+// This is deliberate conservatism: relying on a poll this loop doesn't own
+// would leave the loop naked if a later transform (full unroll, LoopDeletion,
+// vectorization, the second-pass run) removes that sub-loop poll. Keeping each
+// loop self-covered also keeps the coverage invariant locally verifiable.
```
taofengliu: Is this scenario likely to occur in practice?

llvm/lib/Transforms/Jeandle/SafepointElimination.cpp
```diff
@@ -0,0 +1,1851 @@
...
+// (Loops are processed innermost-first, so a sub-loop's polls are already
+// finalized here — the concern is downstream passes, not this one.)
+bool keepOneLoopPoll(Loop &L, LoopInfo &LI, DominatorTree &DT,
```
taofengliu: A call maybe a safepoint.

taofengliu (approved):

PR #49, opened 2026-02-12
feat: Insert G1GC barriers

wazlyqa:
## Related issue(s):

issue #48 

## What this PR does / why we need it:
This PR implements G1GC barrier insertion in the LLVM JIT compiler. It cooprates with https://github.com/jeandle/jeandle-jdk/pull/367

- The pass inserts calls to pre_barrier and post_barrier wrapper functions, which internally select the appropriate barrier based on GC type.
- Pre-barrier receives the derived pointer for correct old value loading.
- Post-barrier receives both the base pointer and the stored value for g1_post_barrier cross-region check.

PR #77, opened 2026-06-17
feat: Support CompressedOops/CompressedClassPointers

wazlyqa:
## Related issue(s):

https://github.com/jeandle/jeandle-jdk/issues/49

Support CompressedOops


rcjhd:
I feel the main risk point is still the JDK's parsing of the StackMap.

The current solution seems better than the old approach to me, because it provides stronger guarantees around ordering correctness.

However, the current approach does have its drawbacks — 
**it introduces fairly invasive modifications to LLVM, changing the semantics/protocol of the StackMap's GCPair.**
But if we keep it tightly coupled with the Jeandle frontend, it should be manageable. Other than that, the metadata footprint is a bit larger.

PR #78, opened 2026-06-18
feat: Support compressedoops

wazlyqa:
## Related issue(s):

https://github.com/jeandle/jeandle-jdk/issues/49

Support CompressedOops


llvm/lib/Transforms/Jeandle/JeandleNarrowOopMarker.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "llvm/IR/Jeandle/Metadata.h"
+#include "llvm/IR/Module.h"
+#include "llvm/IR/Statepoint.h"
+
+using namespace llvm;
+
+namespace {
+
+constexpr uint64_t NarrowOopMarkerType = 6;
+constexpr uint64_t HotSpotTNarrowOop = 16;
```
taofengliu: We need to move the implementation of DeoptValueEncoding to the LLVM side. There is already another PR working on this.
https://github.com/jeandle/jeandle-llvm/pull/72

llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -28556,6 +28556,7 @@ Register X86TargetLowering::getRegisterByName(const char* RegName, LLT VT,
                      .Case("rsp", X86::RSP)
                      .Case("ebp", X86::EBP)
                      .Case("rbp", X86::RBP)
+                     .Case("r12", X86::R12)
```
taofengliu, line 28559: Why we need to add this?

taofengliu (commented):
Do we need an optimization pass to eliminate paired encode/decode operations? Because LLVM's existing optimizations do not necessarily eliminate paired addrspacecast instructions as reversible, we need to confirm this.

llvm/lib/Target/X86/X86ISelLowering.cpp
wazlyqa: On HotSpot x86_64, the compressed-oop heap base is kept in `r12`. When `decode_heap_oop` needs the heap base, Jeandle emits IR like:

    %base = call i64 @llvm.read_register(metadata !"r12")

  During instruction selection, LLVM calls `X86TargetLowering::getRegisterByName("r12", ...)`. Without a mapping from the string `"r12"` to the physical register `X86::R12`, lowering this intrinsic can fail or generate invalid code.

llvm/include/llvm/IR/Jeandle/JeandleUtils.hpp
```diff
@@ -36,12 +36,24 @@ enum HotspotBasicType {
 };
 
 /// Returns true if \p Ty is a pointer in the Java heap address space,
-/// i.e., it represents a Java object reference (oop).
-inline bool isJavaOopType(Type *Ty) {
+/// i.e., it represents a uncompressed Java object reference (oop).
+inline bool isOopType(Type *Ty) {
```
taofengliu:
```suggestion
inline bool isWideOopType(Type *Ty) {
```

llvm/lib/Transforms/Jeandle/ExpandNarrowOopCast.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+} // end anonymous namespace
+
+PreservedAnalyses ExpandNarrowOopCast::run(Function &F,
+                                          FunctionAnalysisManager &) {
+  Module *M = F.getParent();
+
+  // Avoid scanning functions that were not compiled in compressed-oops mode.
+  if (!F.hasFnAttribute(jeandle::Attribute::UseCompressedOops))
+    return PreservedAnalyses::all();
```
taofengliu: Maybe we need to handle JavaOps, too. We can use the attribute of compiled Java method to see if we need to run this pass.

llvm/lib/Transforms/Jeandle/JeandleNarrowOopMarker.cpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+} // end anonymous namespace
+
+PreservedAnalyses JeandleNarrowOopMarker::run(Function &F,
+                                              FunctionAnalysisManager &) {
+  Module *M = F.getParent();
+
+  if (!F.hasFnAttribute(jeandle::Attribute::UseCompressedOops))
+    return PreservedAnalyses::all();
```
taofengliu, line 46: We only need to process the compiled Java method.

llvm/lib/Transforms/Jeandle/JeandleNarrowOopMarker.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+constexpr uint64_t NarrowOopMarkerType = 7;
+constexpr uint64_t HotSpotTNarrowOop = 16;
+
+uint64_t getNarrowOopMarkerEncoding() {
+  // Keep this layout in sync with HotSpot DeoptValueEncoding::encode(). 
+  // The marker has no slot index, so index is zero and the basic type 
+  // records T_NARROWOOP.
+  return (NarrowOopMarkerType << 16) | HotSpotTNarrowOop;
+}
```
taofengliu, line 37: Add a todo here to remind us update it when DeoptValueEncoding is moved to LLVM side and BasicType mirror is added.

llvm/lib/Transforms/Jeandle/JeandleNarrowOopMarker.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+  for (CallBase *CB : Statepoints) {
+    auto *SP = cast<GCStatepointInst>(CB);
+
+    SmallVector<Value *, 16> DeoptInputs;
+    if (auto Bundle = SP->getOperandBundle(LLVMContext::OB_deopt)) {
+      for (const Use &U : Bundle->Inputs)
+        DeoptInputs.push_back(U.get());
+    }
+    bool NeedSyntheticBci = DeoptInputs.empty();
```
taofengliu, line 67: Can DeoptInputs be empty? `JeandleVMState::deopt_args` always push bci as deopt bundle even if there is no locals and stacks. 
Can we assert `!DeoptInputs.empty()`?

llvm/lib/Transforms/Jeandle/JeandleNarrowOopMarker.cpp
wazlyqa: Non-gc-leaf call stubs may also go through statepoint rewriting, and their stackmap GC pairs may contain narrow oops.

llvm/lib/Transforms/Jeandle/JeandleNarrowOopMarker.cpp
wazlyqa: JeandleNarrowOopMarker may also need to handle non-gc-leaf call stubs. `JeandleCallVM::generate_call_VM()` does not add a deopt operand bundle. So for stubs, DeoptInputs can be empty, and the synthetic BCI path is still needed

taofengliu: Add an assertion here: `(DeoptInputs.empty() && !JavaMethodCompilation) || !DeoptInputs.empty()`

taofengliu (approved):

PR #80, opened 2026-07-07
enhance: move JeandleInlineDriver after JavaOperationLower0 and add more passes in runRootInstSimplify

taofengliu:
Move JeandleInlineDriver after JavaOperationLower<0> then we can add passes which rely on JavaOperationLower<0> to runRootInstSimplify, making the runRootInstSimplify more powerful.

### DIFF
--- a/llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
+++ b/llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
@@ -237,8 +237,8 @@ PreservedAnalyses JeandleInlineDriver::run(Module &M,
       // Lower phase-0 JavaOp call sites exposed by the previous inline round's
       // inlining.
       PreservedAnalyses JavaOpLowerPA = JavaOperationLower(0).run(M, MAM);
-      updateDriverPreservedAnalyses(M, MAM, DriverPA, std::move(JavaOpLowerPA));
       Changed |= !JavaOpLowerPA.areAllPreserved();
+      updateDriverPreservedAnalyses(M, MAM, DriverPA, std::move(JavaOpLowerPA));
 
       // Per-round cleanup.
       PreservedAnalyses CleanupPA = runRootInstSimplify(M, MAM);
--- a/llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
+++ b/llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
@@ -169,11 +169,8 @@ static PreservedAnalyses runRootInstSimplify(Module &M,
   FPM.addPass(InstCombinePass());
   FPM.addPass(SimplifyCFGPass());
   FPM.addPass(ADCEPass());
+
   PreservedAnalyses RootPA = FPM.run(*RootFunction, FAM);
-  // FPM.run already invalidated root's analyses per-pass internally, so
-  // RootPA is consumed only for the Changed flag; the combined PA returned
-  // below relies on that per-pass invalidation plus JavaOpLower's eager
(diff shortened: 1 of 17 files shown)


llvm/lib/Transforms/Jeandle/JeandleInlineDriver.cpp
```diff
@@ -189,33 +210,47 @@ PreservedAnalyses JeandleInlineDriver::run(Module &M,
...
+
+    if (RoundChanged) {
+      updateDriverPreservedAnalyses(M, MAM, DriverPA,
+                                    std::move(InlineResult.PA));
+
+      // Lower phase-0 JavaOp call sites exposed by the previous inline round's
+      // inlining.
+      PreservedAnalyses JavaOpLowerPA = JavaOperationLower(0).run(M, MAM);
+      updateDriverPreservedAnalyses(M, MAM, DriverPA, std::move(JavaOpLowerPA));
+      Changed |= !JavaOpLowerPA.areAllPreserved();
```
minchao1024, line 240: Maybe compute `Changed` before moving `JavaOpLowerPA` into `updateDriverPreservedAnalyses`; querying it after `std::move` relies on the moved-from state.
```suggestion
      PreservedAnalyses JavaOpLowerPA = JavaOperationLower(0).run(M, MAM);
      Changed |= !JavaOpLowerPA.areAllPreserved();
      updateDriverPreservedAnalyses(M, MAM, DriverPA, std::move(JavaOpLowerPA));
```

llvm/test/Jeandle/java-operation-deletion.ll
```diff
@@ -0,0 +1,45 @@
+; RUN: opt -S -passes="java-operation-lower<phase=0>,java-operation-deletion" %s 2>&1 | FileCheck %s
+
+; Verify JavaOperationDeletion:
+;   - erases fully-lowered (user-empty) phase-0 JavaOps (gone0, gone1),
+;   - strips them from @llvm.used and removes the now-empty @llvm.used global,
+;   - leaves a JavaOp that still has a real caller (kept0, phase 1, not driven by
+;     phase=0 lowering) in place.
```
minchao1024: Maybe clarify the deletion strategy here. The comment says the `JavaOperationDeletion` pass runs after all `JavaOperationLower` phases, but this test still checks deletion right after phase 0, which looks like the old layered-deletion behavior.

llvm/test/Jeandle/java-operation-deletion.ll
```diff
@@ -0,0 +1,45 @@
...
+}
+
+; gone0 is phase=0, called once by root. phase=0 lowering inlines it into root;
+; it then has no callers (only @llvm.used). Deletion strips @llvm.used and erases.
+define i32 @gone0(i32 %x, ptr addrspace(1) %p) #1 {
+  ret i32 %x
+}
+
+; gone1 is phase=0, called by gone0. phase=0 lowering transitively inlines gone0
+; into gone1; gone1 itself has no callers. Deletion erases it.
```
minchao1024:
```suggestion
; gone1 is phase=0 and calls gone0. phase=0 lowering transitively inlines gone0
; into gone1; gone1 itself has no callers. Deletion erases it.
```

llvm/test/Jeandle/java-operation-deletion.ll
taofengliu: It's just a test to verify JavaOperationDeletion can delete all unused JavaOp. In the real Jeandle pipeline, we run JavaOperationDeletion after the last JavaOperationLower, so it can delete all JavaOps.

minchao1024 (approved):

Issue #81, opened 2026-07-11
Incorrect constant field folding for mixed oop sources


minchao1024:
**Describe the bug**
Jeandle may incorrectly fold a field load when the base oop can come from both a constant oop handle and a runtime oop value.

**Minimal reproduce steps**
```
java -XX:+UseJeandleCompiler -Xbatch -XX:-TieredCompilation -XX:CompileThreshold=1000 -XX:CompileCommand=compileonly,TestCCF::test -XX:-JeandleUseProfile TestCCF
CompileCommand: compileonly TestCCF.test bool compileonly = true
Exception in thread "main" java.lang.RuntimeException: test failed.
        at TestCCF.main(TestCCF.java:11)
```
```
java -XX:-UseJeandleCompiler -Xbatch -XX:-TieredCompilation -XX:CompileThreshold=1000 -XX:CompileCommand=compileonly,TestCCF::test -XX:-JeandleUseProfile TestCCF
CompileCommand: compileonly TestCCF.test bool compileonly = true
test passed.
```
```
// TestCCF.java
class TestCCF {
    public static boolean test(String a, String b, boolean c, boolean d) {
        String obj = c ? a : b;
        String res = d ? obj : "HelloWorld";
        return res.isEmpty();
    }
    public static void main(String[] args) {
        int count = 100000;
        while(count-- > 0) {
            if (!test("", "", true, true)) {
                throw new RuntimeException("test passed.");
            }
        }
        System.out.println("test failed.");
    }
}
```

**Expected behavior**
The program should complete normally.

PR #82, opened 2026-07-11
fix: treat oop arguments as opaque in constant field folding

minchao1024:
## Related issue(s):

issue #81

Seed oop-typed function arguments as Bottom in the ConstOop lattice. Arguments
are opaque producers in the current function context, so relying only on the
fallback path can leave argument-derived forwarders at Top and let them be
incorrectly merged with a constant oop.

This prevents mixed PHI chains such as:
```
  %arg.obj = phi [ %arg0, ... ], [ %arg1, ... ]
  %const.obj = load @oop_handle_Const
  %merged = phi [ %const.obj, ... ], [ %arg.obj, ... ]
  %value = load (%merged + field_offset)
```
from being folded as if %merged were always %const.obj.

Also add a regression test where an oop argument flows through PHIs and is
merged with an oop_handle constant. The field load must remain a load instead
of being folded to the constant oop's field value.


taofengliu (approved):
Nice finding!

Issue #85, opened 2026-07-21
Fix random failue on langtools jtreg cases


Cruise20:
Reproducible case: `test/langtools/jdk/jshell/ToolRetainTest.java`

PR #86, opened 2026-07-21
fix: Fold constant addrspacecast of a null

Cruise20:
## Related issue(s):

issue #85


wazlyqa:
Maybe we do not need to change ExpandNarrowOopCast for this.

I think we can avoid generating this IR in the first place. All Jeandle AS1 <-> AS3 oop casts should go through a helper, e.g. `createCompressedOopCast/createNarrowOopCast`. The helper can special-case `ConstantPointerNull` and directly return a null pointer in the destination address space, instead of calling `IRBuilder::CreateAddrSpaceCast` and producing a `ConstantExpr` addrspacecast(null).

This keeps `ExpandNarrowOopCast` focused on lowering real `AddrSpaceCastInsts`, and makes the null behavior explicit at the Jeandle cast creation site.

taofengliu (approved):

PR #87, opened 2026-07-24
fix: typeIntersect drops Exact on equal-klass intersection

taofengliu:
## Problem
  jeandle::typeIntersect silently discards one operand's Exact flag when the two constraints share the same positive 

```
klass:
     
  typeIntersect({Klass: Animal, Exact: false}, {Klass: Animal, Exact: true})
    → currently: {Klass: Animal, Exact: false}   ❌
    → expected:  {Klass: Animal, Exact: true}    ✓
```

## Root cause

  typeIntersect selects the more specific positive klass using the reflexive IsSubtype query (JavaType.cpp):

```
  if (CB->IsSubtype(A.Klass, B.Klass)) {   // reflexive: TRUE when A.Klass == B.Klass
    Result.Klass = A.Klass;
    Result.Exact = A.Exact;                 // ← ignores B.Exact
  } else if (CB->IsSubtype(B.Klass, A.Klass)) {
    Result.Klass = B.Klass;
    Result.Exact = B.Exact;
  }
```

  IsSubtype is reflexive — K->is_subtype_of(K) returns true (HotSpot Klass::is_subtype_of). So when A.Klass == B.Klass, the first branch is always taken and Result.Exact copies only A.Exact;
  B.Exact is lost.

  Why ||: intersection means both constraints hold simultaneously, and Exact=true is the stricter claim ("the value is exactly this class, not a subclass"). If either operand is exact, the
  intersection is exact. This is the dual of typeUnion, which correctly uses A.Exact && B.Exact for the equal case (a union is exact only when both sources are exact).

## Impact

  A precision bug, not a miscompile — it errs toward the looser, sound type — but it causes downstream consumers to miss optimizations. CHADevirtualization forwards ReceiverType.Exact into
  GetCHAOptInfo (CHADevirtualization.cpp); a dropped Exact weakens or defeats devirtualization.

## Fix

  Handle the equal-klass case explicitly before the subtype checks (this also avoids a VM callback in the common case):

```
  if (A.isKnown() && B.isKnown()) {
    const VMCallbacks *CB = getVMCallbacks();
    assert(CB && CB->IsSubtype && "VMCallbacks must be set");
    if (A.Klass == B.Klass) {
      // Equal klass: Exact is conjunctive — the stricter (exact) claim wins.
      Result.Klass = A.Klass;
      Result.Exact = A.Exact || B.Exact;
    } else if (CB->IsSubtype(A.Klass, B.Klass)) {
      Result.Klass = A.Klass;
      Result.Exact = A.Exact;
    } else if (CB->IsSubtype(B.Klass, A.Klass)) {
      Result.Klass = B.Klass;
      Result.Exact = B.Exact;
    }
  }
```

  The strict-subtype branches are unchanged and already correct: the more-specific klass determines Exact; a supertype-side Exact=true would be contradictory/dead and is correctly ignored.

### DIFF
--- a/llvm/lib/IR/Jeandle/JavaType.cpp
+++ b/llvm/lib/IR/Jeandle/JavaType.cpp
@@ -285,7 +285,15 @@ JavaType jeandle::typeIntersect(JavaType A, JavaType B) {
   if (A.isKnown() && B.isKnown()) {
     const VMCallbacks *CB = getVMCallbacks();
     assert(CB && CB->IsSubtype && "VMCallbacks must be set");
-    if (CB->IsSubtype(A.Klass, B.Klass)) {
+    if (A.Klass == B.Klass) {
+      // Equal klass: Exact is conjunctive. Exact=true is the stricter claim
+      // ("value is exactly this class"), so if either operand is exact the
+      // intersection is exact. IsSubtype is reflexive, so without this case
+      // the first branch below would win and silently drop the other side's
+      // Exact. (Dual of typeUnion's A.Exact && B.Exact for the equal case.)
+      Result.Klass = A.Klass;
+      Result.Exact = A.Exact || B.Exact;
+    } else if (CB->IsSubtype(A.Klass, B.Klass)) {
       Result.Klass = A.Klass;
       Result.Exact = A.Exact;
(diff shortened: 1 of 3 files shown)


minchao1024 (approved):
Looks good.

PR #90, opened 2026-07-27
feat: Support cha for MethodHandle intrinsic.

stevenJOJO129:
## Related issue(s):

[issue #468 in jeandle-jdk](https://github.com/jeandle/jeandle-jdk/issues/468)

1. Support MethodHandle's intrinsic CHA optimization. Make MethodHandle invoke inline possible in constant target situation.


llvm/lib/Transforms/Jeandle/CHADevirtualization.cpp
```diff
@@ -158,13 +436,17 @@ PreservedAnalyses CHADevirtualization::run(Function &F,
   Module *M = F.getParent();
   if (!M->getNamedMetadata(jeandle::Metadata::JavaMethodCompilation))
     return PreservedAnalyses::all();
+  if (F.hasAvailableExternallyLinkage())
+    return PreservedAnalyses::all();
```
taofengliu, line 488: Use `isRootJavaMethodFunction` in jeandle-llvm/llvm/include/llvm/IR/Jeandle/JeandleUtils.h

llvm/lib/IR/Jeandle/JavaType.cpp
```diff
@@ -1084,6 +1096,23 @@ static JavaType getJavaTypeImpl(Value *V, DominatorTree &DT,
...
+    if (isJavaTypeAssumeCall(CB) && Context) {
+      JavaType Input =
+          getJavaTypeImpl(CB->getArgOperand(0), DT, Context, Visited, DestBB);
+      JavaType Assumed = getCallRetJavaType(CB);
+      JavaType Result = typeIntersect(Input, Assumed);
+      JavaType Sharpened = sharpenFromDominators(V, Context, DT, DestBB);
+      Result = typeIntersect(Result, Sharpened);
+      return Result;
+    }
+  }
```
taofengliu: The queried value may not be a direct call to jeandle.assume_java_type; it could instead be a phi node, a select instruction, or some other value derived from a call to jeandle.assume_java_type. 

If we encounter such cases, we will not use the input to of jeandle.assume_java_type to refine the type in a context-sensitive type query, which is problematic. Given the current usage of call jeandle.assume_java_type, I think it is better not to handle this case at all, whether the queried value is a direct call to jeandle.assume_java_type or a phi, select, or other instruction derived from it. This will not affect the effectiveness of the current optimizations.

llvm/lib/Transforms/Jeandle/CHADevirtualization.cpp
```diff
@@ -77,25 +86,282 @@ void updateStaticOptVirtualCallAttrs(InvokeInst &CB, int PatchSize) {
...
+      .DefaultUnreachable("Should not reach here.");
+}
+
+InvokeInst *optimizeMhIntrinsic(InvokeInst &CB, Function &F, DominatorTree &DT,
+                                DomTreeUpdater &DTU,
+                                const jeandle::VMCallbacks &Callbacks,
+                                uintptr_t Caller,
+                                const StringRef &IntrinsicName) {
+  assert(IntrinsicName != "_invokeBasic" &&
+         "_invokeBaisc is treated in normal path.");
```
taofengliu:
```suggestion
         "_invokeBasic is treated in normal path.");
```

taofengliu (commented):
Break the code into sections and add some comments.

taofengliu (commented):
Break the code into sections and add some comments.

llvm/lib/IR/Jeandle/JavaType.cpp
stevenJOJO129: Delete related codes, remain unchanged.

llvm/lib/Transforms/Jeandle/CHADevirtualization.cpp
stevenJOJO129: changed.

llvm/include/llvm/IR/Jeandle/Attributes.h
```diff
@@ -46,7 +46,13 @@ class Attribute {
 
   static constexpr const char *DeclaredHolder = "declared-holder";
 
+  static constexpr const char *MhIntrinsicName = "mh-intrinsic-name";
+
   static constexpr const char *MonomorphicTarget = "monomorphic-target";
+
+  /// Marks a call argument that remains live for the HotSpot runtime ABI even
+  /// when the visible LLVM callee does not use the corresponding parameter.
+  static constexpr const char *RuntimeLive = "runtime-live";
```
taofengliu: Why we need this? It has no reader.

llvm/include/llvm/IR/Jeandle/JeandleUtils.h
```diff
@@ -18,12 +18,15 @@
...
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Type.h"
+#include "llvm/Support/Debug.h"
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/raw_ostream.h"
 
 #include <optional>
 #include <string>
 
+#define DEBUG_TYPE "jeandle-utils"
```
taofengliu: Don't define debug type in a header file. This will lead to redefined macros.

llvm/include/llvm/IR/Jeandle/Attributes.h
stevenJOJO129: In new version code, this is used by DAE pass to avoid receiver marked as poison.

llvm/include/llvm/IR/Jeandle/JeandleUtils.h
stevenJOJO129: Delete this line.

llvm/lib/Transforms/Jeandle/CHADevirtualization.cpp
```diff
@@ -77,25 +78,330 @@ void updateStaticOptVirtualCallAttrs(InvokeInst &CB, int PatchSize) {
...
+    DestKind = jeandle::StaticCall;
+    PatchSize =
+        getPatchSize(CB.getModule(), jeandle::Metadata::StaticCallPatchSize);
+  }
+
+  // Do not commit any speculative type assumptions unless all arguments were
+  // compatible and the VM accepted the new call-site representation.
+  if (!NarrowSuccess ||
+      !Callbacks.UpdateCallSite(static_cast<int64_t>(Id), DestKind,
+                                true, CHAOptInfo.Method)) {
```
taofengliu:
```suggestion
      !Callbacks.UpdateCallSite(static_cast<int64_t>(Id), DestKind, true,
                                CHAOptInfo.Method)) {
```

taofengliu (dismissed):

taofengliu (approved):

PR #92, opened 2026-07-30
feat: support structured VM callback values

xiejf2020:

Extend Jeandle VM callbacks with structured, replayable values.

This adds `Tuple` for statically shaped heterogeneous values and `Array` for dynamically sized homogeneous values. Both are recursively composable from existing callback value types. The callback log now records tuples as `(...)` and arrays as `[...]`, with escaped string literals, so record/replay can preserve the full value tree.

As an initial migration, `GetConstantFieldValue` and `GetConstantFieldInfo` are consolidated into `GetConstantField`, which returns `(BasicType, rawValue)`. This keeps related constant-field information atomic in record/replay and demonstrates the intended tuple-return callback pattern.

The change also adds codec, log round-trip, and record/replay coverage for structured values.


llvm/lib/Transforms/Jeandle/CHADevirtualization.cpp
```diff
@@ -58,6 +59,49 @@ using jeandle::JavaType;
 
 namespace {
 
+// Named view of the tuple returned across the replayable VM callback boundary.
+// ConstraintOrHolder bit 0 distinguishes MethodHandle target information from
+// regular Java invoke information.
+struct CHAOptInfo {
```
taofengliu: We can keep this in the header file and make jeandle-jdk's `jeandleVMCallback.cpp` use this, avoiding discrepancies that may arise from repeated definitions

xiejf2020: fixed

taofengliu (approved):

PR #95, opened 2026-08-04
enhance: add constant folding for oop, mirror, and Klass metadata

rcjhd:
## Related issue(s):

issue https://github.com/jeandle/jeandle-jdk/issues/574

Enhances LLVM constant field folding to resolve:

  - Constant oop to Klass
  - Constant Class mirror to Klass
  - Constant Klass to layout_helper

  This enables further constant propagation and downstream optimizations that depend on precise HotSpot type metadata.

### DIFF
--- a/llvm/lib/Transforms/Jeandle/ConstantFieldFolding.cpp
+++ b/llvm/lib/Transforms/Jeandle/ConstantFieldFolding.cpp
@@ -314,20 +314,6 @@ matchFieldLoad(LoadInst *LI, const DenseMap<Value *, int> &ConstOops,
   return FieldLoadMatch{LI, ImmediateGEP, *BaseId, OffsetVal};
 }
 
-std::optional<uintptr_t> getConstantKlassPointer(Value *V) {
-  V = V->stripPointerCasts();
-  if (auto *CE = dyn_cast<ConstantExpr>(V)) {
-    if (CE->getOpcode() != Instruction::IntToPtr || CE->getNumOperands() != 1)
-      return std::nullopt;
-    if (auto *CI = dyn_cast<ConstantInt>(CE->getOperand(0)))
-      return static_cast<uintptr_t>(CI->getZExtValue());
-    return std::nullopt;
-  }
-  if (auto *CI = dyn_cast<ConstantInt>(V))
-    return static_cast<uintptr_t>(CI->getZExtValue());
-  return std::nullopt;
-}
-
 Constant *klassPointerConstant(LLVMContext &Ctx, Type *Ty, uintptr_t Klass) {
   if (Klass == 0 || !Ty->isPointerTy())
     return nullptr;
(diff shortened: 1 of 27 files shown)


llvm/lib/Transforms/Jeandle/ConstantFieldFolding.cpp
```diff
@@ -302,6 +314,192 @@ matchFieldLoad(LoadInst *LI, const DenseMap<Value *, int> &ConstOops,
   return FieldLoadMatch{LI, ImmediateGEP, *BaseId, OffsetVal};
 }
 
+std::optional<uintptr_t> getConstantKlassPointer(Value *V) {
```
taofengliu: Delete this function. We can use `jeandle::extractKlassConstant` int JavaType.h

llvm/lib/Transforms/Jeandle/ConstantFieldFolding.cpp
```diff
@@ -302,6 +314,192 @@ matchFieldLoad(LoadInst *LI, const DenseMap<Value *, int> &ConstOops,
...
+      klassPointerConstant(CI->getContext(), CI->getType(), Klass);
+  if (!KlassConstant)
+    return false;
+
+  CI->replaceAllUsesWith(KlassConstant);
+  CI->eraseFromParent();
+  return true;
+}
+
+bool foldLoadMirrorKlassCall(CallInst *CI, const jeandle::VMCallbacks &CB,
```
taofengliu, line 429: Maybe we can use type information to fold more.

llvm/lib/Transforms/Jeandle/ConstantFieldFolding.cpp
```diff
@@ -302,6 +314,192 @@ matchFieldLoad(LoadInst *LI, const DenseMap<Value *, int> &ConstOops,
...
+
+  // The frontend emits a null check before Object.getClass. Require LLVM to
+  // prove that the receiver is non-null at this call site as well, so this
+  // fold cannot accidentally remove the Java-required NullPointerException.
+  SimplifyQuery SQ(DL, &DT, nullptr, CI);
+  if (!isKnownNonZero(CI->getArgOperand(0), SQ))
+    return false;
+
+  jeandle::JavaType ReceiverType =
+      jeandle::getJavaType(CI->getArgOperand(0), &DT, CI);
```
taofengliu: Like `foldLoadKlassCall`, we can queryu const oops first.

llvm/lib/Transforms/Jeandle/ConstantFieldFolding.cpp
rcjhd: `JavaType` describes the runtime class of the `java.lang.Class` mirror itself, not the Klass represented by that mirror. Therefore, `foldLoadMirrorKlassCall` cannot safely use ordinary type information, it must rely on the constant mirror identity and `GetMirrorKlass`. Supporting more cases would require separate represented-Klass metadata.

llvm/lib/Transforms/Jeandle/ConstantFieldFolding.cpp
rcjhd: Done

llvm/lib/Transforms/Jeandle/ConstantFieldFolding.cpp
rcjhd: Done

taofengliu (approved):

PR #96, opened 2026-08-05
feat: Partial Escape Analysis

taofengliu:
## Related issue
https://github.com/jeandle/jeandle-jdk/issues/438

## What does this PR do
Implemented Partial Escape Analysis for Jeandle.

Unlike the original PEA paper, we do not move partially escaping allocations to their escape points, because doing so prevents us from constructing the deoptimization operand bundles correctly. As a result, this particular optimization is disabled: a partially escaping allocation still retains its original memory allocation operation. However, operations before the escape point—such as barriers, monitor operations, and field accesses—are still optimized correctly.


llvm/lib/Analysis/Jeandle/PartialEscapeAnalysis.cpp
```diff
@@ -0,0 +1,9736 @@
...
+    return nullptr;
+  if (VBits == 0 || (VBits % 8) != 0)
+    return nullptr;
+
+  // Same bit width → BitCast (Float↔Int, Half↔i16, etc.).
+  if (VBits == LBits) {
+    if (!CastInst::isBitCastable(VTy, LoadTy))
+      return nullptr;
+    Instruction *Cast =
+        CastInst::Create(Instruction::BitCast, V, LoadTy, "pea.coerce",
```
rcjhd, line 2718: Even though this cast is unparented, CastInst::Create immediately adds an operand use to V. The same applies to line 2241, 2305, 2385 also. As a result, merely requesting or caching PartialEscapeAnalysis changes the Value use-lists until the PEAResult is destroyed.

  This breaks the expected analysis/transform separation and may cause another analysis or pass to observe artificial uses, affecting use_empty(), one-use checks, DCE decisions, or similar logic, especially if this analysis is queried without immediately running PartialEscapeTransform.

  We can store a recipe instead, such as the opcode, source value, destination type, and debug location, and let PartialEscapeTransform create the actual instruction, rather than owning temporary IR objects.

llvm/lib/Jeandle/Pipeline.cpp
```diff
@@ -68,21 +106,134 @@ ModulePassManager Pipeline::buildJeandlePipeline(PassBuilder &PB,
...
+      CanonicalizeLPM.addPass(LoopDeletionPass());
+      PrePEAHighTier.addPass(
+          createFunctionToLoopPassAdaptor(std::move(CanonicalizeLPM)));
+    }
+    // Aggressive full-unroll-only: OptLevel=3 raises the cost threshold to
+    // unroll-threshold-aggressive (300, boostable to 400%). Partial,
+    // runtime, and upper-bound unrolling are disabled; peeling stays at the
+    // default policy (it peels a few iterations off loops that resist full
+    // unrolling, exposing straight-line allocations to PEA).
+    PrePEAHighTier.addPass(LoopUnrollPass(
```
rcjhd, line 215: Enabling PEA currently applies aggressive O3 full unrolling, with up to 128 iterations, followed by non-trivial loop unswitching, InstCombine, GVN, and SimplifyCFG to every eligible Java method.

This is potentially expensive for a JIT compiler in terms of compilation latency and generated code size. Maybe we can gate the cluster on the presence of PEA candidate allocations pattern and adding method-size, loop-size, and maximum-growth limits.

Issue #63, opened 2026-04-28
Extend type query interface to support interfaces


taofengliu:
**The feature you'd like**
Jeandle's type query interfaces don't support multiple interfaces sharpening, because in the type query result, only one type is preserved. 
For example, Jeandle cannot eliminate the inner type check in Test::test:
```
public class Test {
    public static void main(String[] args) {
        System.out.println(test(new TestClass()));
    }


    static boolean test(Object obj) {
        if (obj instanceof TestInterface && obj instanceof TestInterface2) {
            if (obj instanceof TestInterface) {
                return true;
            }
        }
        return false;
    }

}

interface TestInterface {
    void test();
}

interface TestInterface2 {
    void test2();
}

class TestClass implements TestInterface, TestInterface2 {
    @Override
    public void test() {
        System.out.println("test");
    }
    @Override
    public void test2() {
        System.out.println("test2");
    }
}
```

In the current implementation, Jeandle will try to sharpen the type of obj by ```obj instanceof TestInterface && obj instanceof TestInterface2```, and only preserve the most specific (most-child-class) type. But the two interfaces are not child class of each other.

**Implementation you've considered**
Extend the JavaType to store interface information collected from type checks.

PR #98, opened 2026-08-18
feat: support multi klass in javatype

NeverRaR:
## Related issue(s):

issue #63 

support multiple interfaces sharpening for java type query.


llvm/include/llvm/IR/Jeandle/JavaType.h
```diff
@@ -37,8 +37,12 @@ namespace jeandle {
...
 /// non-null oop.
 struct JavaType {
-  /// The Klass pointer (from HotSpot JVM). 0 means unknown.
-  uintptr_t Klass = 0;
+  /// The possible Klass pointers (from HotSpot JVM). empty means unknown.
+  /// Only contains at most one non-interface type, and the elements should be
+  /// types that are not subtype of each other;
+  /// Only the most specific (lowwermost) classes are stored;
+  /// more general supertype are implied.
+  SmallDenseSet<uintptr_t, 2> Klasses;
```
taofengliu: I suggest using a separate collection to store interfaces, while keeping the original single variable for non-interface types.

taofengliu (commented):
We need to maintain all the interfaces implemented by the type during the JavaType computation, rather than just those checked via instanceof.

A motivating example:
```
Object x = flag ? new Integer() : new Long();   // Integer and Long                                                                                                                    
x instanceof Comparable                    // must be true
```

taofengliu (approved):

PR #101, opened 2026-08-20
feat: support profile-guided virtual call devirtualization

lfeng14:
## Related issue(s):

issue #

Adds profile-guided devirtualization for Jeandle `invokevirtual` and `invokeinterface` calls. HotSpot reads receiver profiles and resolves concrete targets, while LLVM emits receiver guards and rewrites the selected paths to direct calls.

```text
virtual call
  |
  v
Can CHA determine the target?
  | yes
  +------> CHA guard + direct call
  |
  no
  v
Read receiver profile
  |
  +------> monomorphic
  |           receiver0 -> direct call
  |           miss      -> uncommon trap or virtual call
  |
  +------> bimorphic
  |           receiver0 -> direct call target0
  |           receiver1 -> direct call target1
  |           miss      -> uncommon trap or virtual call
  |
  +------> major receiver
  |           receiver0 -> direct call
  |           other     -> virtual call
  |
  +------> no reliable prediction
              keep the virtual call
```

The miss path uses an uncommon trap until the C2-style trap or recompilation limit is reached, after which it falls back to the original virtual call.

## Main changes
- Select and validate monomorphic, bimorphic, and major-receiver targets from mature profiles.
- Recover the call-site context from LLVM IR and deopt scopes, following CHA; CallSiteInfo remains responsible only for runtime metadata and statepoint tracking.
- Preserve receiver type information after rewriting guarded calls so later type-based optimizations can still use it.
- Preserve class-unloading and accessor information, validate profile counts, and update JDK call-site metadata before rewriting LLVM IR.
- Add JeandleUseProfiledVirtualCallDevirtualization to control the optimization.

## Correctness
- Include `ciMethod` identity in LLVM symbols to distinguish same-named methods from different class loaders.

## Performance
Compared profile devirtualization ON and OFF on 25 Renaissance benchmarks using the release build. Each configuration ran five times on the same NUMA node, with the last three runs averaged. In the reported run, 21 benchmarks improved and four regressed; the geometric-mean improvement was 10.86% and total runtime decreased by 6.16%. The individual regressions were not stable across follow-up runs, and the earlier 8.42% regression reported for `philosophers` could not be reproduced.


llvm/lib/Transforms/Jeandle/JeandleTransformUtils.cpp
```diff
@@ -201,50 +305,43 @@ int getCurrentDeoptBCI(const CallBase &CB) {
...
+            .encode();
+    if (Encoding->getZExtValue() != MethodEncoding)
+      return 0;
+    return static_cast<uintptr_t>(Method->getZExtValue());
+  };
+
+  // Legacy inlinee layout: method marker, method, bci, bci.
+  if (uintptr_t Method =
+          DecodeMethod(Scope.BCIPairStart - 2, Scope.BCIPairStart - 1))
+    return Method;
```
taofengliu: Maybe we don't need this

llvm/lib/Transforms/Jeandle/JeandleTransformUtils.cpp
```diff
@@ -20,10 +20,114 @@
...
+  uintptr_t ExistingMethod = 0;
+  return F->getFunctionType() == Type &&
+         getFunctionJavaMethod(*F, ExistingMethod) && ExistingMethod == Method;
+}
+
+Function *getOrInsertJavaMethodFunction(Module &M, StringRef Name,
+                                        FunctionType *Type, uintptr_t Method,
+                                        bool IsAccessor) {
+  if (!canGetOrInsertJavaMethodFunction(M, Name, Type, Method))
+    return nullptr;
```
taofengliu, line 83: Caller in CHA should check nullptr

llvm/lib/Transforms/Jeandle/JeandleTransformUtils.cpp
lfeng14: fixed. The legacy inlinee fallback is no longer needed and has been removed. Inlinee scopes now use only the current method-marker and should_reexecute layout.

llvm/lib/IR/Jeandle/JavaType.cpp
```diff
@@ -486,8 +486,37 @@ struct TraceResult {
...
+  uintptr_t Klass = 0;
+
+  bool matched() const { return Klass != 0; }
+};
+
 } // anonymous namespace
 
+/// Match `jeandle.check_exact_klass(ExpectedKlass, ActualKlass)`, where
+/// `ActualKlass` is loaded directly from `QueryObj` by `jeandle.load_klass`.
+static ExactKlassGuard traceExactKlassGuard(Value *Cond, Value *QueryObj) {
```
taofengliu: We should handle the case where load_klass appears as the first argument here.

llvm/lib/IR/Jeandle/JavaType.cpp
```diff
@@ -486,8 +486,37 @@ struct TraceResult {
   }
 };
 
+/// An exact receiver-class guard generated by profile devirtualization.
+struct ExactKlassGuard {
+  uintptr_t Klass = 0;
+
+  bool matched() const { return Klass != 0; }
+};
```
taofengliu: Just use Klass as return type. Remove this

taofengliu (approved):

PR #111, opened 2026-09-02
fix: handle invokeBasic target signature mismatch

xiejf2020:

Fix `_invokeBasic` CHA devirtualization for adapted MethodHandles whose wrapper signature differs from the target method signature. The optimizer now resolves the concrete MethodHandle target, validates the target argument types, rebuilds the direct call with the target’s actual function type, and safely handles discarded return values for `void` wrappers. Ordinary `_linkTo*` CHA handling remains unchanged. This keeps the generated LLVM call type consistent with the real Java target and avoids correctness failures caused by reusing the outer wrapper signature.


xiejf2020:
`jdk-dependent` needed

taofengliu (approved):

Issue #106, opened 2026-08-27
Support late inlining in the Jeandle inliner


minchao1024:
**The feature you'd like**

Add late-inlining support to the Jeandle inliner.

The current inlining callback only allows the VM to immediately accept or reject an inline candidate. The VM should also be able to defer a candidate until eager inlining reaches a fixed point and relevant optimization passes have run.

A deferred call should then be reconsidered using the VM's late-inlining policy. The inliner should also handle the VM's node-count cutoff decision and prevent terminally failed late-inline candidates from being retried indefinitely.

**Implementation you've considered**

Replace `IsOkToInline` with a new `GetInlineDecision` callback that receives the inline scope, BCI, callee method, and whether the call is being considered during late inlining.

The callback would return one of the following decisions:

- `Deny`
- `InlineNow`
- `InlineLater`
- `HitNodeCountCutoff`

Calls returning `InlineLater` would be marked with scheduling metadata. The inline driver would use a one-way transition from eager inlining to late inlining, with a dedicated hook for optimization passes that should run immediately before each late-inline round.

The inliner would also:

- Preserve inline-scope information across rounds.
- Process newly exposed calls depth-first while preserving IR order.
- Mark terminal late-inline failures as `noinline`.
- Correctly invalidate analyses when call-site attributes are modified.
- Remove temporary scheduling metadata before leaving the pass.
- Update callback recording/replay and the related LLVM tests.

**Additional context**

The numeric values of the inline decisions must remain stable because they are persisted in callback logs and replay files.

This change requires corresponding updates to the JDK-side callback implementation so that the LLVM and JVM callback ABIs remain consistent.

The implementation should preserve the community branch's existing PEA callbacks, replay scope, and additional `RecoverTypeInfo`/ADCE cleanup passes.

The current implementation has been validated with all inline tests passing, as well as the complete `llvm/test/Jeandle` test suite with no failures.

PR #113, opened 2026-09-03
feat: add late inline support to Jeandle inliner

minchao1024:
## Related issue(s):

issue #106 

Replace IsOkToInline with GetInlineDecision to support eager, delayed,
denied, and node-count-cutoff decisions.

Add one-way eager-to-late scheduling and depth-first inline processing.
Preserve inline scope metadata across rounds, prevent failed late-inline
candidates from being retried, and correctly invalidate analyses when
adding noinline attributes.

Update callback replay logs and add tests for successful and failed
late inlining.

### DIFF
--- a/llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
+++ b/llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
@@ -572,6 +572,11 @@ InlineRoundResult JeandleInliner::runInlineRound(
       continue;
     }
 
+    // Jeandle's inline policy consumes JVM profile through VM callbacks, while
+    // imported inlinees have no LLVM function entry counts. Branch/switch
+    // profile metadata is cloned with the IR, and caller analyses are
+    // invalidated after each inline, so profile redistribution and BFI
+    // maintenance are unnecessary here.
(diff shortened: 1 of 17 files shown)


llvm/lib/Transforms/Jeandle/JeandleInliner.cpp
```diff
@@ -437,19 +494,44 @@ InlineRoundResult JeandleInliner::runInlineRound(
...
+        VC->GetInlineDecision(InlineScopeID, BCI, CalleeMethod, IsLateInline);
+    auto Decision = static_cast<jeandle::JeandleInlineDecision>(RawDecision);
+    if (Decision == jeandle::JeandleInlineDecision::HitNodeCountCutoff) {
+      logInlineEvent("node-count-cutoff", ScopeCaller, BCI, Callee,
+                     InlineScopeID, ThreadID, InlineScopes);
+      HitNodeCountCutoff = true;
+      if (InLateInlinePhase) {
+        HasLateInlineCandidates = false;
+      }
+      break;
```
taofengliu, line 513: We cannot break here because every callee's node count cut off is different. We should try to inline other callees.

PR #114, opened 2026-09-03
enhance: JavaType system

taofengliu:

Motivation:
- OSR type recovery: during OSR compilation, a local variable of reference type is a bare load (with no metadata/attributes), and the only source of type information is the checkcast emitted by the frontend at the OSR entry; checkcast is a phase-0 JavaOp, expanded at the start of the pipeline, and its expanded form is subsequently normalized away, so dominator-chain analysis cannot recover it
- Direct use after merging multiple predecessors with the same check (diamond shape: both branches must pass check_instanceof(K, o) before reaching the merge point, and o is used directly at the merge point)
- Nested merge-of-merge (the inner merge block is a predecessor of the outer merge block)

This PR uses CFG based type sharpening to replace the current dominator tree based type sharpening.


NeverRaR:
LGTM

Cruise20 (approved):

